PR #20, opened 2024-03-20
APP-4292 viam-server increase stop timeout

ale7714:
https://viam.atlassian.net/browse/APP-4292 

viam-server has an overall stop timeout of 90 seconds. We observed that the sigkill sent by the agent at 60 seconds can interfere with the clean shutdown process of viam-server. For this reason, I think the timeout of the agent should be higher than the viam-server one. With this change the timeout will be 2minutes.


subsystems/viamserver/viamserver.go
```diff
@@ -171,7 +171,7 @@ func (s *viamServer) Stop(ctx context.Context) error {
 		s.logger.Error(err)
 	}
 
-	if s.waitForExit(ctx, stopTimeout/2) {
+	if s.waitForExit(ctx, stopTimeout) {
```
Otterverse, line 174: I think this is the wrong line to modify. The timeout above this one (line 163) is the wait before sending KILL. That's probably the one to modify. This one is just waiting for everything to be dead which SHOULD be only second or so at most, as kill is kernel-only, just yanks the rug. No chance for the program to delay things on its own.

cheukt:
we need to update the timeout at https://github.com/viamrobotics/agent/blob/main/manager.go#L212 as well

manager.go
```diff
@@ -209,7 +209,7 @@ func (m *Manager) SubsystemHealthChecks(ctx context.Context) {
 
 // CloseAll stops all subsystems and closes the cloud connection.
 func (m *Manager) CloseAll() {
-	ctx, cancelFunc := context.WithTimeout(context.Background(), time.Minute)
+	ctx, cancelFunc := context.WithTimeout(context.Background(), time.Minute*2)
```
cheukt, line 212: nit: leave comment as to why we chose the time we did, and maybe make this a constant as well

subsystems/viamserver/viamserver.go
```diff
@@ -160,7 +160,7 @@ func (s *viamServer) Stop(ctx context.Context) error {
...
```
cheukt, line 163: nit: leave comment as to why we chose the time we did

cheukt (approved):
LGTM with some requests for more comments

subsystems/viamserver/viamserver.go
ale7714: sorry! i literally got confused on lines because they looked almost the same

manager.go
```diff
@@ -209,7 +210,8 @@ func (m *Manager) SubsystemHealthChecks(ctx context.Context) {
 
 // CloseAll stops all subsystems and closes the cloud connection.
 func (m *Manager) CloseAll() {
-	ctx, cancelFunc := context.WithTimeout(context.Background(), time.Minute)
+	// 2 min timeout. It must be higher than viam-server shutdown timeout of 90 secs.
```
Otterverse, line 213: Systemd has a stop timeout of 90s by default. Thus, when doing an actual shutdown, this time limit applies more strongly than the "longer" one for viam-server itself to avoid systemd force-killing the agent itself (and considering it a failure.) If you're raising this, you'll need to also raise the limit (including some margin) in the systemd service file (and thus, in the preinstall script as well.)

subsystems/viamserver/viamserver.go
```diff
@@ -160,7 +160,8 @@ func (s *viamServer) Stop(ctx context.Context) error {
 		s.logger.Error(err)
 	}
 
-	if s.waitForExit(ctx, stopTimeout/2) {
+	// 2 min timeout. It must be higher than viam-server shutdown timeout of 90 secs.
```
Otterverse, line 163: Now that you've changed the TOTAL stopTimeout time (in the manager) to 2 minutes, this needs to be shorter than that by a decent margin, to allow other subsystems time to shutdown as well (they're stopped sequentially.)

Otterverse (changes requested):
Some more notes... unfortunately all these timeout values I picked were chosen with specific reasons and interact a lot with each other.

manager.go
Otterverse: Just realized, we already have TimeoutSec set to 120seconds, so just need to raise that to probably 240s or so

ale7714:
This change updates the chain of timeouts more broadly to:

- Systemd timeout to 240 secs 
- Viam Agent CloseAll timeout to 180 sec
- Viam server subsystem 
    - sig term timeout 120 sec 
    - sig kill to 10 sec
- All other subsystems
   - sig term 30 secs
   - sig kill 10 sec

benjirewis (approved):
Nice! Thanks 🧑‍🔧 .

subsystems/viamagent/viam-agent.service
```diff
@@ -8,7 +8,7 @@ Type=exec
 Restart=always
 RestartSec=5
 User=root
-TimeoutSec=120
+TimeoutSec=240
```
Otterverse, line 11: This should be modified to match in preinstall.sh (though it's not critical, as it'll update the first time it gets online anyway.)

Otterverse (changes requested):
One minor thing, otherwise looking good!

Oh, and make sure to run lint before pushing. I think a linting failure snuck in from the last PR. (Can't wait to get that automated.)

Otterverse (approved):
LGTM!

PR #21, opened 2024-03-20
[APP-4862] Improve lockfile, add faststart mode, add synfs calls, cleanup

Otterverse:
Test build (for pin_url): http://packages.viam.com/temp/viam-agent-vfastsync-aarch64

This PR is attempting to improve/cleanup the several hotfixes implemented in the past two weeks, along with additional cleanup.

Key areas:
- Adds syncfs() calls around all key file writes related to updates [APP-4862]
    - This should help prevent zero-length file issues when power is lost shortly after updates (e.g. during brief power-up testing in manufacturing.)
- Switches lockfiles back to the viam directory
    - Relying on a systemd run directory isn't fully safe. It's far too easy for someone to start things OUTSIDE of systemd, thus defeating the purpose of the lockfile entirely. (Systemd itself already stops duplicates internally.) 
    - I believe I cover the likely cause of the issue. If a hard crash or power cut happens, the lockfile is left intact. On systems without PID randomization, PIDs are generated sequentially. This means on reboot, there's likely to be a NEW process with the same PID as the one viam-agent had on the previous boot, so the lock appears to be valid and blocks.
- Adds faststart mode
    - This makes bypassing the self-update (and network checks) opt-in via an environment variable, CLI arg, or subsystem attribute in the cloud config
        - Set `"fast_start": true` in the viam-server subsystem attributes
    - Normal users likely WANT an update on boot if their machine has been off/unused for a while
- Move systemd service file to /usr/local/lib/systemd/system instead of the /etc/ location
    - This allows user overrides (with full file edit) to work.
    - Includes a one-time removal of the old /etc/ located file
    - THIS NEEDS TESTING ON VARIOUS SYSTEMS
        - Not 100% certain every OS we try to load this on has that path in its systemd config
        - Can verify with: `systemd-path systemd-search-system-unit | grep -o /usr/local/lib/systemd/system`
- Lots of cleanup
    - Especially trying to make error message texts more consistent

[APP-4862]: https://viam.atlassian.net/browse/APP-4862?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


cmd/viam-agent/main.go
```diff
@@ -176,20 +167,46 @@ func main() {
...
 	}
 
-	// Start viam server as soon as possible. Then, start other subsystems and check for updates
-	if err := manager.StartSubsystem(ctx, viamserver.SubsysName); err != nil {
-		if errors.Is(err, agent.ErrSubsystemDisabled) {
-			globalLogger.Warn("viam-server subsystem disabled, please manually update /etc/viam.json and connect to internet")
-		} else {
-			globalLogger.Errorf("could not start viam-server subsystem: %s", err)
+	// if faststart is set, skip updates and start viam-server immediately, the proceed as normal
+	if opts.Fast {
```
ale7714, line 171: we have to let Pete know that tenibot will need to enable this

cmd/viam-agent/main.go
```diff
@@ -248,3 +265,55 @@ func exitIfError(err error) {
 		globalLogger.Fatal(err)
 	}
 }
+
+func getLock() (lockfile.Lockfile, error) {
```
ale7714, line 269: nice! 👏

subsystems/viamagent/viamagent.go
```diff
@@ -24,8 +24,8 @@ func init() {
 
 const (
 	subsysName      = "viam-agent"
-	serviceFileDir  = "/etc/systemd/system"
-	serviceFilePath = "/etc/systemd/system/viam-agent.service"
+	serviceFileDir  = "/usr/local/lib/systemd/system"
```
ale7714, line 27: what would be the downside of not doing this? is this mostly a best practice cleanup?

Otterverse: So Pete found that, when he tried to remove the "After=" line from the service file with `systemctl edit --full viam-agent` it REPLACES the file in /etc/ as that's meant to be the user/admin override location. The downside is that, on next update, we then try to "repair" that file with a new copy, losing his customizations.

The short version is /etc is for user/owner config, whereas /usr/lib/ is for vendoer/OS installed services (stuff controlled by the OS's package manager), and /usr/local/lib/ is where stuff in between (like us) goes. And the other locations all can be ultimate overridden by the user in /etc.

cmd/viam-agent/main.go
Otterverse: Yeah, I definitely will let him know before this goes live, and will ask him to test and validate it still works for his use case before then too. I don't think they're shipping quite as urgently soon as canyon runner though, so hopefully it'll be a minor thing add. (The updates in the readme show exactly what to do to enable this.)

Otterverse: Had a thought... this should be a config option in the cloud (as an attribute), not an ENV var. More work to implement it, but makes more sense that way, and allows control of this at fleet-level with upcoming fragments.

subsystems/viamagent/viamagent.go
ale7714: Ohh that's a good observation

README.md
```diff
@@ -36,6 +36,11 @@ To remove the agent and all viam configuration completely, run the following scr
 sudo /bin/sh -c "$(curl -fsSL https://storage.googleapis.com/packages.viam.com/apps/viam-agent/uninstall.sh)"
 ```
 
+## FastStart Mode
+This bypasses the normal network/online wait and update checks during inital startup, and executes viam-server as quickly as possible. Useful if you have a device that often starts when offline or on a slow connection, and having the latest version immediately after start isn't required. Note that normal, periodic update checks will continue to run afterwards. This only affects initial startup sequencing.
+
+To use it, set `"fast_start": true` in the attributes for the viam-server subsystem. Alternately, set `VIAM_AGENT_FASTSTART=1` in your environment.
```
ale7714, line 42: how are we going to manage this for tennibot? are we going to manipulate their config?

install.sh
```diff
@@ -243,7 +243,7 @@ main() {
 
 	uninstall_old_service
 
-	if [ -f /etc/systemd/system/viam-agent.service ]; then
+	if [ -f /etc/systemd/system/viam-agent.service ] || [ -f /usr/local/lib/systemd/system/viam-agent.service ]; then
```
ale7714, line 246: this is for another distro?

ale7714: nvm I just kept reading and remember that we're moving where the file is

subsystems/viamagent/viamagent.go
```diff
@@ -96,72 +101,112 @@ func GetRevision() string {
...
 		}
 		if err := os.Symlink(curPath, expectedPath); err != nil {
-			return errors.Wrapf(err, "cannot install symlink at %s", expectedPath)
+			return errw.Wrapf(err, "installing symlink at %s", expectedPath)
 		}
 	}
 
+	if inSystemdPath(serviceFileDir, logger) {
+		// one-time, remove old /etc based service file when moving to new /usr/local location
+		removeOldServiceFile(logger)
```
ale7714, line 140: when we do this we have already written the new service file to disk?

utils.go
```diff
@@ -254,5 +254,18 @@ func ForceSymlink(orig, symlink string) error {
 	if err != nil {
 		return errw.Wrap(err, "symlinking file")
 	}
-	return nil
+
+	return SyncFS(symlink)
+}
+
+func SyncFS(syncPath string) (errRet error) {
```
ale7714, line 261: 👍  

Not for this PR but it feels like we should move this into a go-utils so both the agent and viam-server use the same code.

ale7714:
Overall looks good to me! i just have a few questions mostly how are we going to handle tennibot

ale7714 (approved):

subsystems/viamagent/viamagent.go
Otterverse: Actually, no... just realized that breaks thing if anyone ever tries to use overwrites (the main reason for moving this.) The function itself only runs when the new file isn't yet in place (so it never runs again in the future.) I'll figure something else out.

ale7714 (approved):
thanks for updating

manager.go
```diff
@@ -342,7 +342,7 @@ func (m *Manager) saveCachedConfig(cfg map[string]*pb.DeviceSubsystemConfig) err
 		return err
 	}
 	//nolint:gosec
-	return os.WriteFile(cacheFilePath, cacheData, 0o644)
+	return errors.Join(os.WriteFile(cacheFilePath, cacheData, 0o644), SyncFS(cacheFilePath))
```
dgottlieb, line 345: To the degree these files need to persist, the directory needs to be fsynced as well.

subsystem.go
```diff
@@ -215,7 +215,7 @@ func (s *AgentSubsystem) saveCache() error {
 		return err
 	}
 	//nolint:gosec
-	return os.WriteFile(cacheFilePath, cacheData, 0o644)
+	return errors.Join(os.WriteFile(cacheFilePath, cacheData, 0o644), SyncFS(cacheFilePath))
```
dgottlieb, line 218: Same fsync directory remark

subsystem.go
```diff
@@ -610,5 +610,5 @@ func (is *InternalSubsystem) Update(ctx context.Context, cfg *pb.DeviceSubsystem
 
 	// If attribute changes, restart after writing the new config file.
 	//nolint:gosec
-	return true, os.WriteFile(is.cfgPath, jsonBytes, 0o644)
+	return true, errors.Join(os.WriteFile(is.cfgPath, jsonBytes, 0o644), SyncFS(is.cfgPath))
```
dgottlieb, line 613: Same fsync directory remark

utils.go
```diff
@@ -92,7 +93,7 @@ func DownloadFile(ctx context.Context, rawURL string) (outPath string, errRet er
 			return "", err
 		}
 		defer func() {
-			errRet = errors.Join(errRet, outfd.Close())
+			errRet = errors.Join(errRet, outfd.Close(), SyncFS(outPath))
```
dgottlieb, line 96: Similarly, to the degree this file needs to be removed, fsync the directory after calling `os.Remove`.

utils.go
```diff
@@ -133,7 +134,7 @@ func DownloadFile(ctx context.Context, rawURL string) (outPath string, errRet er
 		return "", err
 	}
 	defer func() {
-		errRet = errors.Join(errRet, out.Close())
+		errRet = errors.Join(errRet, out.Close(), SyncFS(out.Name()))
```
dgottlieb, line 137: Fsync after `Remove`

dgottlieb (commented):
I used the code changes to guide where we do FS operations. I think the "bottom of the PR" was most important as that deals with the end result file.

PR #25, opened 2024-06-07
Fix panic on errors.As()

Otterverse:
Quick fix for a panic I hit during some manual testing. Don't think it's critical enough to warrant its own release, but didn't want to lose/forget the fix.

Also, tweaked the makefile so it won't accidentally tag binaries with stable versions when uncommitted changes are present.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,8 +6,10 @@ else ifeq ($(GOARCH),arm64)
 LINUX_ARCH = aarch64
 endif
 
-GIT_REVISION = $(shell git rev-parse HEAD | tr -d '\n')
-TAG_VERSION ?= $(shell git tag --points-at | sort -Vr | head -n1 | cut -c2-)
+ifeq ($(shell git status -s),)
+	TAG_VERSION ?= $(shell git tag --points-at | sort -Vr | head -n1 | cut -c2-)
+	GIT_REVISION = $(shell git rev-parse HEAD | tr -d '\n')
+endif
 ifeq ($(TAG_VERSION),)
 PATH_VERSION = custom
 else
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -6,8 +6,10 @@ else ifeq ($(GOARCH),arm64)
 LINUX_ARCH = aarch64
 endif
 
-GIT_REVISION = $(shell git rev-parse HEAD | tr -d '\n')
-TAG_VERSION ?= $(shell git tag --points-at | sort -Vr | head -n1 | cut -c2-)
+ifeq ($(shell git status -s),)
+	TAG_VERSION ?= $(shell git tag --points-at | sort -Vr | head -n1 | cut -c2-)
```
abe-winter, line 10: there is also a `git describe --tags` command which creates a tag/sha/dirty-aware version string

may or may not be useful for your purposes

abe-winter (approved):
I don't understand why the old way would panic, but lgtm if it works for you

Otterverse:
> I don't understand why the old way would panic, but lgtm if it works for you

For posterity: `errors.As panics if target is not a non-nil pointer to either a type that implements error, or to any interface type.` For the ExitError type, the Error() is implemented on a pointer, so this ends up needing to be a pointer to a pointer.

This was what pointed me (heh) to that solution: https://github.com/golang/go/issues/51120#issuecomment-1034231219

Makefile
Otterverse: Good to know, but it seems to just point to the last commit then, which means uncommitted changes still get the identical tag to the latest commit, so not _quite_ useful for this case at the moment.

PR #26, opened 2024-06-10
APP-5248 use NetAppender in agent

abe-winter:
## Checklist
- [x] wait for https://github.com/viamrobotics/rdk/pull/4091
- [x] make sure viam-agent non datestamped logs are still going to stdout
## What changed
- replace golog + zap with rdk/logging everywhere
- do NetAppender startup in Manager.dial
- `globalNetAppender` for MatchingLogger + subsystem log parse / upload
## Why
Make agent logs visible in webapp.
## Reviewer questions
- ~I checked that this doesn't double-log, but why doesn't it? the matching logger? I thought that rdk logs showed up in journalctl. is that via subprocess inheriting stderr, not via the logger?~ -> I get it now
## Follow-up changes
- add netappender delayed start [APP-5330](https://viam.atlassian.net/browse/APP-5330) which will let us undo the uglier changes here
- When the device is offline, `NetAppender.Close` takes a long time (10 second timeout). Agent + RDK are both doing this, so offline shutoff now takes 20+ seconds. Ticketed [RSDK-7866](https://viam.atlassian.net/browse/RSDK-7866)
- should cloud logging be configurable? (i.e. turn it off). where?
- rate limiting

[RSDK-7866]: https://viam.atlassian.net/browse/RSDK-7866?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[APP-5330]: https://viam.atlassian.net/browse/APP-5330?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


cmd/viam-agent/main.go
```diff
@@ -164,6 +164,16 @@ func main() {
 		}
 	}
 
+	netAppender, err := manager.NetAppender(ctx)
```
Otterverse, line 167: This likely needs to be after the "fast start" section just below. Fast start is supposed to work entirely offline and as quickly as possible (specifically for tennibot) and this leads to dialing (and timeouts) vastly slowing that offline behavior I believe.

Might be best for this to live within the dial() function? E.g. it'll set up the network appender whenever anything first dials, regardless of source.

manager.go
```diff
@@ -95,6 +95,25 @@ func (m *Manager) LoadConfig(cfgPath string) error {
...
+	}
+	return logging.NewNetAppender(
+		&logging.CloudConfig{
+			AppAddress: m.cloudAddr,
+			ID:         m.partID,
+			Secret:     m.cloudSecret,
+		},
+		m.conn,
+	)
+}
```
Otterverse, line 115: As noted in the other comment, I think this could be rolled into dial() to remove some repitition here. E.g. the chance for valid cloud configs is already done by LoadConfig(), so no need to duplicate that again here (once it's in an internal function that can't get called out of order.)

subsystems/viamserver/viamserver.go
```diff
@@ -267,9 +267,9 @@ func (s *viamServer) Update(ctx context.Context, cfg *pb.DeviceSubsystemConfig,
 	return false, nil
 }
 
-func NewSubsystem(ctx context.Context, logger *zap.SugaredLogger, updateConf *pb.DeviceSubsystemConfig) (subsystems.Subsystem, error) {
+func NewSubsystem(ctx context.Context, logger logging.Logger, updateConf *pb.DeviceSubsystemConfig) (subsystems.Subsystem, error) {
 	setFastStart(updateConf)
-	return agent.NewAgentSubsystem(ctx, SubsysName, logger, &viamServer{logger: logger.Named(SubsysName)})
+	return agent.NewAgentSubsystem(ctx, SubsysName, logger, &viamServer{logger: logger.Sublogger(SubsysName)})
```
Otterverse, line 272: How does the output of something with "Sublogger()" differ from the original Named() call?

Otterverse (changes requested):
Looking good over all, and glad this is turning out to be mostly straight forward so far. I've put one main request inline (move some stuff into dial() so it simplifes and works with fast_start), but also wanted to address your question:

> I checked that this doesn't double-log, but why doesn't it? the matching logger? I thought that rdk logs showed up in journalctl. is that via subprocess inheriting stderr, not via the logger?

Look at the logger.Write() (logging.go:60). Two things here. First, matches can be marked as "masking" meaning the output ONLY goes to the match channel, and is hidden from normal logging (this is currently only used for health checks.)

The second bit is that it looks for a specific timestamp format coming in. If it's already a "log" line with a specific timestamp format in it, then it just writes directly to stdout/stderr, instead of the logger. (And actually, if something doesn't have a timestamp, it logs it with a warning about it being "unstructured output") The intent was to avoid double-timestamps on logger-formatted lines coming from subsystems, but this will need to be reworked somehow to make sure these ARE now passed into the regular logger, in order to make it to the cloud, but we don't want to "nest" them with double-timestamps. I'm not sure if the best solution is to somehow parse and forward the original (more accurate) timestamp, or to just strip it all and re-log the original message, while letting the new logger timestamp as it sees fit. I expect milliseconds of difference at most in normal cases, but haven't fully thought through if there are edge cases.

subsystems/viamserver/viamserver.go
abe-winter: it's dotted, so it will be `viam-agent.viam-server`

`.Named()` produces a zap logger which I think has a copy not a reference; if you call this line first, then do AddAppender later, the appender doesn't show up on the zap logger from `.Named()`

cmd/viam-agent/main.go
abe-winter: felt initially weird to put it in dial, then realized it *has* to be dial because we don't know when we will come online

done in ec27bc3

should I do anything to protect against a second netappender getting added on reconnect? I think this doesn't happen now, but could see it accidentally happening if we refactor in the future

(for example, do `SetNetAppender` instead of `AddNetAppender`)

manager.go
```diff
@@ -95,6 +96,35 @@ func (m *Manager) LoadConfig(cfgPath string) error {
...
+	}
+	if m.cloudAddr == "" || m.partID == "" || m.cloudSecret == "" {
+		return ErrMissingCloudCreds
+	}
+	netAppender, err := logging.NewNetAppender(
+		&logging.CloudConfig{
+			AppAddress: m.cloudAddr,
+			ID:         m.partID,
+			Secret:     m.cloudSecret,
+		},
```
Otterverse, line 117: Probably harmless (and/or good practice), but normally this should never be used, as we're passing a connection directly, right? I'm a little uncertain if we DO pass creds here, if/when it might use them and re-dial its own separate connection. Do we want that? (Genuinely asking!)

manager.go
```diff
@@ -95,6 +96,35 @@ func (m *Manager) LoadConfig(cfgPath string) error {
 	return nil
 }
 
+var ErrMissingCloudCreds = errors.New("can't create NetAppender without cloud creds")
```
Otterverse, line 99: I'm not sure why this error is special (and gets a variable) when the other two similar checks below just directly `return errors.New(...)`

manager.go
```diff
@@ -376,12 +409,17 @@ func (m *Manager) dial(ctx context.Context) error {
...
+	conn, err := rpc.DialDirectGRPC(ctx, u.Host, m.logger.AsZap(), dialOpts...)
 	if err != nil {
 		return err
 	}
 	m.conn = conn
 	m.client = pb.NewAgentDeviceServiceClient(m.conn)
+
+	// todo: ideally we would create the NetAppender ASAP (so logs are captured) and only *connect* it here.
+	if err := m.attachNetAppender(); err != nil {
+		m.logger.Errorw("error attaching NetAppender", "err", err)
```
Otterverse, line 421: It feels like this should be returning the error here, not just logging it. What's the scenario you're expecting where this is non-fatal? From looking at the code, I think the ONLY possible error this could ever log would be if if it failed to get the hostname (from the os.Hostname() call inside NewNetAppender()) which I'd consider a pretty critical failure. All the others are impossible to hit per my above comment.

manager.go
```diff
@@ -376,12 +409,17 @@ func (m *Manager) dial(ctx context.Context) error {
...
 
-	conn, err := rpc.DialDirectGRPC(ctx, u.Host, m.logger, dialOpts...)
+	conn, err := rpc.DialDirectGRPC(ctx, u.Host, m.logger.AsZap(), dialOpts...)
 	if err != nil {
 		return err
 	}
 	m.conn = conn
 	m.client = pb.NewAgentDeviceServiceClient(m.conn)
+
+	// todo: ideally we would create the NetAppender ASAP (so logs are captured) and only *connect* it here.
```
Otterverse, line 419: Is this something you're working on for this PR? If not, we (and I include myself here) should get in the habit of making all long term TODOs into jira tickets and linking them. E.g. `TODO(RSDK-9999): blah blah blah`

subsystems/viamserver/viamserver.go
Otterverse: That's fine, was just curious what the difference was.

manager.go
```diff
@@ -95,6 +96,35 @@ func (m *Manager) LoadConfig(cfgPath string) error {
...
+func (m *Manager) attachNetAppender() error {
+	if m.conn == nil {
+		return errors.New("NetAppender requires non-null conn")
+	}
+	if m.netAppender != nil {
+		return errors.New("Manager already has non-nil netAppender")
+	}
+	if m.cloudAddr == "" || m.partID == "" || m.cloudSecret == "" {
+		return ErrMissingCloudCreds
+	}
```
Otterverse, line 111: Does this need to be a separate function? When I suggested adding this to dial, I meant to simplify things. At the point this is called (bottom of dial()), all these error checks are impossible to hit, as dial() would've already returned an error.

Otterverse (changes requested):
So some nits inline, and maybe a little bit of confusion about intent when I mentioned putting things in dial().

Big issue is that the (network) logging from subsystems still hasn't been dealt with. Or am I wrong and that's somehow working as intended?

manager.go
abe-winter: this is an unattended device -- if it can possibly start up, it should

os.hostname could fail if procfs is unreadable https://cs.opensource.google/go/go/+/refs/tags/go1.22.4:src/os/sys_linux.go;l=40

(or the implementation of netappender could change such that it returns errors in more cases)

manager.go
abe-winter: done 5312398

manager.go
abe-winter: removed in 5312398

I think I was planning to detect this + fail differently? but not sure, removed

abe-winter:
screenshot of subsystem logs flowing through

Screenshot 2024-06-12 at 10-23-00 my-laptop - Viam

logger.go
```diff
@@ -96,3 +112,72 @@ func (l *MatchingLogger) Write(p []byte) (int, error) {
 
 	return len(p), nil
 }
+
+// parsedLog is a lightweight log structure we parse from subsystem logs.
+// Another approach for capturing logs from subsystems is to pass around
+// LogEntry or opentelemetry structs.
```
Otterverse, line 118: Ugh, this feels so... "wrong" to have to do it this way. Sorry, not saying you did it wrong, just, there should be a cleaner way in our logging system to handle this. A zap-style log line parser just doesn't feel like it should be living directly in the agent code. Also shouldn't be having to manually write to netappender either.

Do you think some of this would be better in the logging package itself? (Along with adding a Write(zapcore.Entry) function?)

logger.go
```diff
@@ -96,3 +112,72 @@ func (l *MatchingLogger) Write(p []byte) (int, error) {
...
+// stripAnsiColorCodes removes color codes from a string so we can use it internally.
+func stripAnsiColorCodes(raw []byte) []byte {
+	return colorCodeRegexp.ReplaceAll(raw, nil)
+}
+
+// entry converts a parsedLog to a zapcore.Entry which can be NetAppender'd.
+func (p parsedLog) entry() zapcore.Entry {
+	level, ok := levels[string(p.level)]
+	if !ok {
+		level = zapcore.DebugLevel
```
Otterverse, line 152: An unknown level means the parsing has gone wrong, doesn't it? Feels like that should be Warn or higher to me so it's not masked.

logger.go
```diff
@@ -96,3 +112,72 @@ func (l *MatchingLogger) Write(p []byte) (int, error) {
...
+
+// getIndexOrNil returns the element at `index` of `arr`, or nil if out of range.
+func getIndexOrNil[T any](arr [][]T, index int) []T {
+	if index < len(arr) {
+		return arr[index]
+	}
+	return nil
+}
+
+func parseLog(line []byte) *parsedLog {
```
Otterverse, line 174: I feel this needs to return an error instead of using the getIndexOrNil() stuff. If we can't parse the line, I think it's important to consider that an error and write the WHOLE line to the log. That is, revert to the "Unstructured Log Output" behavior that previously existed.) Otherwise, a parsing failure means we could lose critical information when it's likely the most critical to know it (e.g. something unexpected is happening.)

Otterverse (changes requested):
Mostly questions, and a couple (opinionated) nits from me. (I'm not a fan of mutable globals :-) )

Main concern is that this still isn't logging early output from viam-server, which means we can't catch failures there, and similarly, log line parsing failures mean potentially losing information too.

Some open questions as well. Feel free to poke me on slack to discuss.

logger.go
abe-winter: yeah agree re kludge

it's not a completely new problem; would imagine stackdriver, splunk, fluentd et al have to deal with this at some point as aggregators. wonder if there are are standard tools for this task

no preference about where it lives. if you want to move it can that be a fast-follow rather than a blocker?

Otterverse: I feel it needs more thought, but its more important to get this out... so yeah, doesn't have to be a blocker on this PR.

logger.go
```diff
@@ -79,8 +90,13 @@ func (l *MatchingLogger) Write(p []byte) (int, error) {
 	}
 
 	// filter out already-timestamped logging from stdout
-	dateRegex := regexp.MustCompile(`^[0-9]{4}-[0-9]{2}-[0-9]{2}T`)
 	if dateRegex.Match(p) {
+		// TODO(RSDK-7895): these lines are sometimes multi-line.
+		if l.upload {
+			if parsed := parseLog(p); parsed.valid() && globalNetAppender != nil {
+				globalNetAppender.Write(parsed.entry(), nil) //nolint:errcheck,gosec
```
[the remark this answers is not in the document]
abe-winter: was concerned that if I logged the error, it would end up in the same function and hang forever on the lock at L87 above

options:
- create a non-netappender logger for this kind of error? it wouldn't be logged to network, but it would end up in journalctl
- delayed call in goroutine with some kind of recursion limit so it won't get into 'an error logging an error' situations

logger.go
```diff
@@ -1,33 +1,44 @@
...
 	"github.com/pkg/errors"
-	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
+	"go.viam.com/rdk/logging"
 )
 
+var dateRegex = regexp.MustCompile(`^[0-9]{4}-[0-9]{2}-[0-9]{2}T`)
+
+// globalNetAppender receives matching logger writes if non-nil.
+var globalNetAppender *logging.NetAppender
```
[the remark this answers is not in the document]
abe-winter: oh you're right this is potentially way simpler than I'm making it

didn't realize the l.logger here was the globalLogger passed down from Manager

rdk logger does the appender stuff itself, so Core().Write() isn't an option, but we could make the `Logger.log(*LogEntry)` method public and get the behavior we need

https://github.com/viamrobotics/rdk/blob/main/logging/impl.go#L157-L165

if we do this, I'd have to replace the `if dateMatched { os.Stdout.Write(p) }` behavior above with the parseLog().entry() stuff -- are you okay with that?

logger.go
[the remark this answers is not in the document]
abe-winter: done in 5c81cbf6

logger.go
```diff
@@ -96,3 +112,72 @@ func (l *MatchingLogger) Write(p []byte) (int, error) {
...
+	"DEBUG":  zapcore.DebugLevel,
+	"INFO":   zapcore.InfoLevel,
+	"WARN":   zapcore.WarnLevel,
+	"ERROR":  zapcore.ErrorLevel,
+	"DPANIC": zapcore.DPanicLevel,
+	"PANIC":  zapcore.PanicLevel,
+	"FATAL":  zapcore.FatalLevel,
+}
+
+var colorCodeRegexp = regexp.MustCompile(`\x1b\[\d+m`)
```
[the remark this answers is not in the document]
abe-winter: done in bc17cb78

subsystems/viamserver/viamserver.go
```diff
@@ -77,8 +77,8 @@ func (s *viamServer) Start(ctx context.Context) error {
 		s.shouldRun = true
 	}
 
-	stdio := agent.NewMatchingLogger(s.logger, false)
-	stderr := agent.NewMatchingLogger(s.logger, true)
+	stdio := agent.NewMatchingLogger(s.logger, false, false)
+	stderr := agent.NewMatchingLogger(s.logger, true, false)
```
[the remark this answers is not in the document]
abe-winter: done in 1190ea58 but if we get the logger.Core().Write() stuff above working, this will change again, so don't look until we've made a design decision

note that if those early logs happen in the pre-online section, they will get missed anyway

logger.go
abe-winter: done in 1cf86b7d

logger.go
abe-winter: done in eb094112 + added tests to make sure the parser fallback behaves as expected

abe-winter:
not waiting on you -- was in a holding pattern until we decided what to do
about Logger.Write PR, now rehabbing

manager.go
abe-winter: if we get to this part of the function, `m.client` is non-nil and a reattempt would short-circuit anyway (in the `m.client != nil` guard above)

I think this gets easier once we do NetAppender delayed start in https://viam.atlassian.net/browse/APP-5330 -- okay to keep as-is for now, then improve when the NetAppender change is done?

abe-winter:
okay I think this now has all desired log types (see screenshot):
1. agent's own logs
2. subsystem logs
3. viam logs appear + are not doubled
4. unstructured errors

this is kind of ugly -- Logger.Write didn't fully solve the problem we wanted it too. I opened a ticket [APP-5330](https://viam.atlassian.net/browse/APP-5330) to add 'delayed start' functionality, which should allow us to go back to the old way + simplify a lot of this

logging screenshot

logger.go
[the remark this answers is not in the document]
Otterverse: Okay, after discussion, fine with this as long as APP-5330 is considered priority (for agent) before whatever next (significant) change we do here.

manager.go
Otterverse: Yeah, again, as long as it's done before we do OTHER new projects in agent (though no rush until then), I'm fine with that (and agree it'll be easier/cleaner.) I just don't want to let tech debt pile up.

abe-winter:
retested after cleanup in 667a0c7. confirmed output still has:
- both types of subsystem: subsystem.go + viamserver.go
- subsystem subprocess
- non-doubled viam-server logs
- unstructured output

log:
```log
6/24/2024, 2:43:58 PM info viam-agent agent/subsystem.go:483 agent-syscfg started
6/24/2024, 2:43:58 PM info agent-syscfg viam-agent-syscfg/main.go:77 agent-syscfg startup complete
6/24/2024, 2:43:53 PM info robot_server impl/local_robot.go:1230 Robot (re)configured
6/24/2024, 2:43:53 PM error viam-agent unstructured output: ALSA lib pulse.c:242:(pulse_connect) PulseAudio: Unable to connect: Connection refused
6/24/2024, 2:43:50 PM info viam-agent viamserver/viamserver.go:76 Starting viam-server 
```

subsystems/viamserver/viamserver.go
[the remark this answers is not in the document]
Otterverse: So this now has the opposite problem. It no longer LOCALLY logs viam-server output at all. I just tested this with a quick build on my own pi.

Local logging needs to always happen (as it'd never be duplicated.) It's only the CLOUD logging where we need to log early startup of viam-server (in case it fails), then stop once it's online and doing its own cloud logging.

logger.go
```diff
@@ -75,24 +99,115 @@ func (l *MatchingLogger) Write(p []byte) (int, error) {
...
+		err := globalNetAppender.Write(entry.Entry, nil)
+		if err != nil {
+			fmt.Printf("error writing to NetAppender %s", err) //nolint:forbidigo
+		}
+	} else if l.uploadAll {
+		// in this case, date matching succeeded and we think this is a parseable log message.
+		// we check uploadAll because some subprocesses have their own netlogger which will
+		// upload structured logs. (But won't upload unmatched logs).
+		entry := parseLog(p).entry()
+		l.logger.Write(&logging.LogEntry{Entry: entry})
```
Otterverse, line 125: This needs to ALWAYS happen, not just when uploadAll is set. The logger is writing the local log too, and that local log needs to always be written.

Otterverse (changes requested):
After discussions, I'm okay with postponing some of the "correctness" things until the rdk logger is improved to allow delays and stuff.

The only remaining blocker IMHO is that viam-server logs are NEVER written locally now. These need to always be logged locally.

manager.go
abe-winter: yup agree, bumped 5330 to high and will submit a netappender PR sometime this week

Otterverse (changes requested):
Cool, global is gone, and what your'e doing it cleaner. Two small comment things inline, but the one block is a nil pointer panic.

manager.go
```diff
@@ -231,6 +259,8 @@ func (m *Manager) CloseAll() {
 	m.connMu.Lock()
 	defer m.connMu.Unlock()
 
+	m.netAppender.Close()
```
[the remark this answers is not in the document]
abe-winter: thanks for catch -- ran into this while removing netappender and did not think about close case

zooming out, would be good to have 'can it start and stop cleanly without network connection' as a CI test case

this is done in 21379349

tested that it can close with wifi disconnected: netappender takes like 20 seconds to fully exit (that's ticketed [here](https://viam.atlassian.net/browse/RSDK-7866)), but no crash
```
2024-06-24T20:19:46.828Z	INFO	netlogger	logging/net_appender.go:233	error logging to network: rpc error: code = DeadlineExceeded desc = context deadline exceeded
2024-06-24T20:19:56.834Z	INFO	netlogger	logging/net_appender.go:233	error logging to network: rpc error: code = DeadlineExceeded desc = context deadline exceeded
2024-06-24T20:20:01.838Z	INFO	netlogger	logging/net_appender.go:233	error logging to network: rpc error: code = DeadlineExceeded desc = context deadline exceeded
```

logger.go
```diff
@@ -1,33 +1,54 @@
...
 
-// MatchingLogger provides a zap logger that also allows sending regex matched lines to a channel.
+// MatchingLogger provides a logger that also allows sending regex matched lines to a channel.
 type MatchingLogger struct {
 	mu           sync.RWMutex
-	logger       *zap.SugaredLogger
+	logger       logging.Logger
 	matchers     map[string]matcher
 	defaultError bool
+	// if uploadAll is false, only send unstructured log lines to globalNetAppender.
```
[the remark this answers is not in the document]
abe-winter: edited in 21379349

manager.go
```diff
@@ -95,6 +96,33 @@ func (m *Manager) LoadConfig(cfgPath string) error {
...
 }
 
+// create and attach a NetAppender.
+func (m *Manager) attachNetAppender() error {
+	if m.conn == nil {
+		return errors.New("NetAppender requires non-null conn")
+	}
+	if m.netAppender != nil {
+		return errors.New("Manager already has non-nil netAppender")
+	}
```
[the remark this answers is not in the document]
abe-winter: done in 21379349

Otterverse (approved):
LGTM! Sorry this turned into such a large amount of back and forth!

PR #30, opened 2024-07-11
RSDK-7888 attach NetAppender earlier

abe-winter:
## What changed
- create net appender immediately after LoadConfig, rather than waiting for successful `dial()`. Do NetAppender.SetConn in dial
- in Manager, use a `logging.CloudConfig` struct instead of the three floating vars. This gives me a clean way to test if config has been loaded when creating the netappender
## Why
The NetAppender needs to start as early as possible or else we lose some logs at the beginning.
## Manual testing
I ran this on my laptop + checked that the RDK log page contains:
- viam-agent logs
- subsystem logs (agent-provisioning, agent-syscfg)
- non-duplicated RDK logs (i.e. agent isn't duping rdk's logger)
- unstructured error output
## Follow-ups
- Currently AddAppender is happening after `Manager.LoadConfig`. Add a `NetAppender.SetConfig` method so this can happen at the very top of main + we can capture all logs (especially provisioning).
- Or ... invest in a fully durable logging approach that can survive restart offline etc


manager.go
```diff
@@ -39,12 +39,10 @@ type Manager struct {
...
 	conn        rpc.ClientConn
 	client      pb.AgentDeviceServiceClient
-	partID      string
-	cloudAddr   string
-	cloudSecret string
+	cloudConfig *logging.CloudConfig
 
 	logger      logging.Logger
-	netAppender *logging.NetAppender
+	NetAppender *logging.NetAppender
```
Otterverse, line 45: I'm a bit lost here... First, why make this public? Second, I can't find where this is set now, since you removed the "attachNetAppender()" function.

manager.go
```diff
@@ -89,38 +87,20 @@ func (m *Manager) LoadConfig(cfgPath string) error {
...
-		},
-		m.conn,
-	)
-	if err != nil {
-		return err
+func (m *Manager) CreateNetAppender() (*logging.NetAppender, error) {
+	if m.cloudConfig == nil {
+		return nil, errors.New("can't create NetAppender before config has been loaded")
 	}
-	m.netAppender = netAppender
```
Otterverse, line 121: Don't you still need this line? Without it, I can't find how/where m.NetAppender is ever actually set.

Otterverse (changes requested):
I think there's a bug, as nothing is ever actually setting m.NetAppender. Details inline.

abe-winter:
total miss by me -- ran something mid-change, tested against the wrong binary, and when I saw correct behavior, assumed I was good. fixed in 1264bdd

Otterverse (approved):
LGTM! As discussed, I'll do some version bumps when RDK releases next week and then publish the actual agent release after that.

PR #32, opened 2024-08-18
Update README.md

npentrel:
As we now have all of this information in the docs, I'd like to remove it here so that we have one source of truth. This helps ensure we make updates in the place most useful to our users.


Otterverse (changes requested):
Note: strong OPINIONS incoming. :-)

I'm REALLY not in favor of removing this. README and in-code documentation are supposed to be the source of truth, as it (should) change exactly when the code itself does, and gets maintained by the author(s) of the code itself, so less is lost in translation. Long format documentation should be based on this in turn.

That said, I've no objection to providing the link to the docs at the top here, and suggesting people go there first even.

npentrel:
The problem with that is that things go out of date because authors update the readme and then do not update the docs or file follow up tickets. Removing the readme forces engineers to do the latter.

Otterverse:
I understand the desire and frustration with things being out of sync, but I genuinely believe this approach would make things worse, not better. Overall, I want to wait for Ale and/or others to weigh in on this, because either way, this probably needs to be a company-wide policy.

In the meantime, I'm going to present the argument for how I view things.

> The problem with that is that things go out of date because authors update the readme and then do not update the docs or file follow up tickets. Removing the readme forces engineers to engage with the docs and keep those updated as the source of truth.

See, I feel it's the opposite. This doesn't force them to do that. It allows them to more easily IGNORE it entirely.

Having in-code documentation means PRs can be rejected if the README is not updated to match changes. Thus it forces engineers to maintain documentation that's always in sync with the code, before the PR even merges. As a side effect, that means it also forces OTHER engineers (the reviewers) to verify the technical correctness of the documentation change as part of PR review. And it's documentation written directly by the engineer making the changes, and at the time they're made, absolutely minimizing confusion. Without that, it's too easy for engineers to simply go "eh, it's a small change, it doesn't REALLY need a docs update" and then a bunch of "small" changes get left undocumented, and things drift.

With a README, even for a small change, PR reviews should always force the engineer to update the README (if needed) before it's approved. Then when closing the task, it's a very simple question if a docs ticket is needed. Did you have to change the README? Then yes, it's needed. The extra bonus is that ideally the docs ticket just points to the PR, and the docs team can see exactly what changed in the README, and thus easily update the docs site. Without that, some weeks later when the docs ticket is picked up, someone from docs has to go hunt down the engineer that made the changes, hope they remember what they were, and then hope that they can be explained correctly and clearly. Because the notes on the docs ticket itself are likely to say something vague (but that seemed sufficient at the time) like "reorganized attribute structure." It's MUCH easier to enforce clarity when it's part of the PR, and not left to a jira ticket or a conversation/discussion that may be weeks later.

Otterverse (approved):
After discussion with @ale7714 the decision has apparently been made to remove READMEs across the board. I stated my case, and now bow out.

PR #33, opened 2024-08-27
configurable `startTimeout` for viam-server subsystem

abe-winter:
## What changed
- make the start timeout for the viam-server subsystem configurable
- incidental -- expand VIAM_AGENT_DEVMODE to include running as non-root. (convenience feature for local testing)
## Why
On slow connections and with large binaries, the default 5 minute timeout isn't enough to download viam-server and modules.
## Example config
(attributes.start_timeout is the new part)
```json
{
  "agent": {
    "viam-server": {
      "release_channel": "stable",
      "pin_version": "",
      "pin_url": "",
      "attributes": {"start_timeout": "1h"},
      "disable_subsystem": false
    }
  },
  "components": [
```
## Testing
Tested on my local machine with the new config value missing and present. When present it logs as expected:
> 2024-08-28T15:12:44.807Z DEBUG   viam-agent      viamserver/viamserver.go:91     parsed duration 10m0s from key start_timeout

### DIFF
--- a/subsystems/viamserver/viamserver.go
+++ b/subsystems/viamserver/viamserver.go
@@ -25,6 +25,7 @@ import (
 )
 
 func init() {
+	globalConfig.Store(&viamServerConfig{startTimeout: defaultStartTimeout})
 	registry.Register(SubsysName, NewSubsystem, DefaultConfig)
 }
 
@@ -47,7 +48,7 @@ var (
 
 	// Set if (cached or cloud) config has the "fast_start" attribute set on the viam-server subsystem.
 	FastStart    atomic.Bool
-	globalConfig = &viamServerConfig{startTimeout: defaultStartTimeout}
+	globalConfig atomic.Pointer[viamServerConfig]
 )
 
 type viamServer struct {
@@ -174,7 +175,7 @@ func (s *viamServer) Start(ctx context.Context) error {
 		return nil
 	case <-ctx.Done():
 		return ctx.Err()
-	case <-time.After(globalConfig.startTimeout):
+	case <-time.After(globalConfig.Load().startTimeout):
 		return errw.New("startup timed out")
 	case <-s.exitChan:
 		return errw.New("startup failed")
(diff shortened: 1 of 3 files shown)


subsystems/viamserver/viamserver.go
```diff
@@ -135,7 +154,7 @@ func (s *viamServer) Start(ctx context.Context) error {
 		return nil
 	case <-ctx.Done():
 		return ctx.Err()
-	case <-time.After(startTimeout):
+	case <-time.After(s.startTimeout()):
```
stevebriskin, line 157: It'd be better to parse all the necessary fields out of the config into a new struct and use those directly. Easier for logging and debugging.
This may be fine for a one-off but this pattern won't scale. Is this a config-reading pattern we use in the rest of the code base?

stevebriskin (changes requested):

subsystems/viamserver/viamserver.go
abe-winter: switch to something more general in 9155ca4

in the codebase as-is we use go-flags for env/cli flags, other than that just proto and json afaik

the golang config library I'm most familiar with is viper -- has defaults + duration parsing, would probably make this PR a purely declarative change. (though our config approach has three layers and may not fit perfectly in a third party library)

Otterverse (changes requested):
A few issues (inline) on this. It's also overlapping with code that exists in agent-provisioning, and as I'm working to merge the two, I'm going to share those functions. How urgent is this? If it can wait a few weeks for the code merge, I'm happy to work this in as a new config option during that (and save you further effort here.) If you need something sooner/urgent until then, I can overlook some things and know I'll refactor it during the merge anyway.

But I have a bigger concern overall... 5+ minutes is already LONG. I don't think just raising the timeout even higher is the right answer here. Do we have tickets to fix this on the viam-server side? Seeing a robot stay offline for more than 5 minutes after hitting restart is gonna cause more problems and confusion for users than just here. Definitely feel the robot should go "live" much earlier, and then reconfigure as module/package downloads finish.

Oh, and FYI, the timeout here is for startup of viam-server itself, until it starts listening on port 8080. It doesn't include the time the AGENT spends downloading viam-server (specifically because I was worried about long downloads messing with the timeout.)

subsystems/viamserver/viamserver.go
```diff
@@ -263,12 +302,14 @@ func (s *viamServer) Update(ctx context.Context, cfg *pb.DeviceSubsystemConfig,
 		s.logger.Info("awaiting user restart to run new viam-server version")
 		s.shouldRun = false
 	}
+	globalConfig = configFromProto(s.logger, cfg)
```
Otterverse, line 305: Technically race-possible. (Shouldn't happen in normal operation, but a lot of fast-fails in a row or something might cause it to happen.) Needs lock or atomic like fast_start uses. Maybe https://pkg.go.dev/sync/atomic#Pointer ?

Otterverse (changes requested):
Looks fine, save one last thing around locking the pointer swap during Update. (mentioned in slack too)

subsystems/viamserver/viamserver.go
abe-winter: done in 8c2219d

Otterverse (approved):
LGTM! As discussed, will refactor this somewhat during the merge-of-subsystems project.

PR #34, opened 2024-08-28
consume _DEVMODE var through opts

abe-winter:
## What changed
Move `VIAM_AGENT_DEVMODE` variable to the `opts` struct parsed by go-flags.
## Why
Visibility / consistency
## New help output
```sh
$ go run ./cmd/viam-agent/ -h
Usage:
  viam-agent runs as a background service and manages updates and the process lifecycle for viam-server.

Application Options:
  -c, --config=   Path to config file (default: /etc/viam.json)
  -d, --debug     Enable debug logging (for agent only) [$VIAM_AGENT_DEBUG]
  -f, --fast      Enable fast start mode [$VIAM_AGENT_FAST_START]
  -h, --help      Show this help message
  -v, --version   Show version
      --install   Install systemd service
      --dev-mode  Allow non-root and non-service [$VIAM_AGENT_DEVMODE]
```

### DIFF
--- a/cmd/viam-agent/main.go
+++ b/cmd/viam-agent/main.go
@@ -39,9 +39,9 @@ func main() {
 	ctx := setupExitSignalHandling()
 
 	var opts struct {
-		Config  string `default:"/etc/viam.json"                            description:"Path to config file" long:"config" short:"c"`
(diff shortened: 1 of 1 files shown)


Otterverse (approved):
LGTM... had been meaning to do that some time ago, then forgot as it never came up (I'm always testing in regular mode to make sure it works.)

PR #36, opened 2024-10-01
Bump goutils

benjirewis:


Otterverse (approved):
LGTM

benjirewis:
Need to test this doesn't break current "get" functionality on the agent. I also may want to increase some network timeouts in this PR if there are any in the agent since network operations behind the bluetooth SOCKS proxy are especially slow. Do you foresee any issue with that @Otterverse ?

Otterverse:
> Need to test this doesn't break current "get" functionality on the agent. I also may want to increase some network timeouts in this PR if there are any in the agent since network operations behind the bluetooth SOCKS proxy are especially slow. Do you foresee any issue with that @Otterverse ?

Yes, potentially. The main one is obviously the download time. I don't remember if there's a "timeout" specifically, but... a really long download can be a really long block at startup, since on a fresh start, updates are downloaded/installed before each subsystem is started. So I expect some work may need to be done to background downloads and not block the normal healthcheck/startup loop. I'm not really sure what that looks like.

That said, bumping goutils alone isn't likely to make current behavior WORSE, so still good to merge on this specific PR... or it can become a bigger one if you want. Just note that I'm changing a lot of this very soon, so may not want to put in the effort on that yet.

benjirewis:
> That said, bumping goutils alone isn't likely to make current behavior WORSE, so still good to merge on this specific PR... or it can become a bigger one if you want. Just note that I'm changing a lot of this very soon, so may not want to put in the effort on that yet.

Sounds good; I won't change any timeouts here and will just bump goutils. I testing locally and the agent still seems to work as expected.

PR #39, opened 2024-11-10
Fix network settings error when gateway empty

ale7714:
* Make network gateway "optional" as there are cases where it may not be needed (local network)
* Add debug log line with parsed network settings
* Skip activating connection if error to add or update


Otterverse (approved):
LGTM

PR #42, opened 2024-11-15
RSDK-9223 Conditionally restart viam-server based on restart_status

benjirewis:
[RSDK-9223](https://viam.atlassian.net/browse/RSDK-9223)

Modifies `Update` to check the `restart_status` HTTP endpoint of the `viam-server` subsystem when an update is available. If that endpoint indicates that an external restart is allowed, immediately restart the `viam-server` instance for an update.

Tested locally with an old `viam-server` (no exposed `restart_status` endpoint) and a new `viam-server` version (exposed `restart_status` endpoint.) Restarts and logs happen as (I) expected.

Will leave in "Draft" until the associated `rdk` PR is merged.

[RSDK-9223]: https://viam.atlassian.net/browse/RSDK-9223?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

cc @viamrobotics/netcode


subsystems/viamserver/viamserver.go
```diff
@@ -294,17 +296,79 @@ func (s *viamServer) HealthCheck(ctx context.Context) (errRet error) {
...
+	var needRestart bool
+
 	if newVersion && s.running {
-		s.logger.Info("awaiting user restart to run new viam-server version")
 		s.shouldRun = false
+
+		// viam-server can be safely restarted even while running if the process
+		// has reported it is safe to do so through its `restart_status` HTTP
+		// endpoint.
+		restartAllowed, err := s.isRestartAllowed(ctx)
```
benjirewis, line 356: Discussed with both of you offline about the only "risk" from the scope:
> Risk: A single config update triggers both a viam-server version update and a reconfiguration, and viam-agent tries to restart viam-server while it is in the middle of a reconfiguration due to a race.
Mitigation: (Avoid) Check the restart_status URL for a {“restart_allowed”: true} field in agent multiple times (4 over 60 seconds) in viam-agent before actually restarting viam-server.

What I have now does _not_ avoid the risk and does _not_ implement the "Mitigation" described above. 

@Otterverse I know we spoke about checking the `restart_status` endpoint once when an update is possible for `viam-server`, waiting N minutes, checking the `restart_status` endpoint again, and, only if both checks had indicated a possible external restart, restarting `viam-server`.

After discussing more offline and in the `rdk` PR with @dgottlieb, we would be in favor of the "dumber" strategy here: check the `restart_status` once when a restart is availabe and restart `viam-server` if it indicates an allowed restart.

Let's call an accidental restart of the `viam-server` while it's reconfiguring a "false positive" and any delay in restarting the `viam-server` when a new version is available and the `viam-server` is not reconfiguring a "false negative." The "N minute" strategy seems to lower false positives and raise false negatives. The "dumber" strategy I have here seems to raise false positives and lower false negatives. I would argue that neither approach eliminates false positives, as I can come up with an antagonistic case where the "N minute" strategy still results in a false positive.

If we _really_ wanted to avoid false positives, we think we would need a scope change and a slightly different protocol here. Something like a) `viam-server` handles its own shutdown for restart or b) `viam-server` returns a duration of time that it can be safely be restarted in `restart_status`. In our opinion, the false positive case is not dangerous enough to warrant that change + addition of complexity.

Let me know what you think; this is much more your codebase than NetCode's. And, thanks for all the help in implementing this.

subsystems/viamserver/viamserver.go
```diff
@@ -294,17 +296,79 @@ func (s *viamServer) HealthCheck(ctx context.Context) (errRet error) {
...
+		if err != nil {
+			return false, err
+		}
+		if restartAllowed {
+			s.logger.Infof("will restart %s to run new version as it has reported allowance of a restart",
```
ale7714, line 361: can we log the current version and the new one?

subsystems/viamserver/viamserver.go
```diff
@@ -294,17 +296,79 @@ func (s *viamServer) HealthCheck(ctx context.Context) (errRet error) {
...
+		//nolint:gosec
+		client := &http.Client{Transport: &http.Transport{TLSClientConfig: &tls.Config{InsecureSkipVerify: true}}}
+
+		resp, err := client.Do(req)
+		if err != nil {
+			return false, errw.Wrapf(err, "checking whether %s allows restart", SubsysName)
+		}
+
+		defer func() {
+			errRet = errors.Join(errRet, resp.Body.Close())
```
dgottlieb, line 323: I know this is a pattern in our codebase, but (I claim) there's no utility in checking the error on a `Close`.

If the whole function worked fine, should a caller not proceed because `Close` was unhappy?

subsystems/viamserver/viamserver.go
```diff
@@ -294,17 +296,79 @@ func (s *viamServer) HealthCheck(ctx context.Context) (errRet error) {
...
+		restartAllowed, err := s.isRestartAllowed(ctx)
+		if err != nil {
+			return false, err
+		}
+		if restartAllowed {
+			s.logger.Infof("will restart %s to run new version as it has reported allowance of a restart",
+				SubsysName)
+			needRestart = true
+		} else {
+			s.logger.Infof("awaiting user restart to run new %s version", SubsysName)
```
dgottlieb, line 365: Is this actually true? I thought this case would mean that the the viam-server returned "don't restart please, I'm reconfiguring".

In which case we'd expect the next `Update` loop to observe that reconfigure it done and the viam-agent would restart as expected. Without user intervention I think?

subsystems/viamserver/viamserver.go
benjirewis: From what I can tell, we do not have access to the current or previous version here. There _is_ a log that will appear right before this one:
```
2024-11-15T17:54:57.796Z	INFO	viam-agent	agent/subsystem.go:357	viam-server updated to 0.50.0
2024-11-15T17:54:57.921Z	INFO	viam-agent	viamserver/viamserver.go:361	will restart viam-server to run new version as it has reported allowance of a restart
```
I could extend that log to say `viam-server updated from [previous version] to 0.50.0` if that sounds good to @Otterverse? Would that be in `s.CacheData.PreviousVersion`?

subsystems/viamserver/viamserver.go
benjirewis: I would agree, was just matching pattern from `HealthCheck` above. Used `utils.UncheckedError` instead here and above, but I don't have a strong preference. Either way, it would be good if the HTTP response `Close` here and above handled their errors identically.

subsystems/viamserver/viamserver.go
benjirewis: Whoops; good catch 😮‍💨 . Changed log message accordingly.

subsystems/viamserver/viamserver.go
Otterverse: So I think it's much more likely to hit false positives without a delay here. When someone clicks "save" in the config page, if they made any changes that would require a reconfig and change the viam-server version, both these things kick off at the same time.

That said, if you feel it's not "dangerous" (it will be a signaled/controlled shutdown via SIGTERM after all) to hit those false positives occasionally, then that's fine by me. Simpler is nice in a lot of ways.

subsystems/viamserver/viamserver.go
```diff
@@ -294,17 +296,81 @@ func (s *viamServer) HealthCheck(ctx context.Context) (errRet error) {
 	return errRet
 }
 
+func (s *viamServer) isRestartAllowed(ctx context.Context) (restartAllowed bool, errRet error) {
```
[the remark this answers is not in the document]
Otterverse: If you don't need/use the error return, it's probably easiest to just log it and forget it (or use utils.UncheckedError() )

subsystems/viamserver/viamserver.go
```diff
@@ -294,17 +297,80 @@ func (s *viamServer) HealthCheck(ctx context.Context) (errRet error) {
...
```
Otterverse, line 300: Should comment that this needs to be called only from inside a lock (since it uses s.checkURL/s.checkURLAlt fields from s.)

subsystems/viamserver/viamserver.go
```diff
@@ -294,17 +297,80 @@ func (s *viamServer) HealthCheck(ctx context.Context) (errRet error) {
...
 	if newVersion && s.running {
-		s.logger.Info("awaiting user restart to run new viam-server version")
 		s.shouldRun = false
+
+		// viam-server can be safely restarted even while running if the process
+		// has reported it is safe to do so through its `restart_status` HTTP
+		// endpoint.
+		restartAllowed, err := s.isRestartAllowed(ctx)
+		if err != nil {
+			return false, err
```
Otterverse, line 359:
```suggestion
			return needRestart, err
```
nit: but if you're gonna have a variable set for this, should use it

subsystems/viamserver/viamserver.go
Otterverse: I believe that's the correct source.

subsystems/viamserver/viamserver.go
```diff
@@ -294,17 +297,80 @@ func (s *viamServer) HealthCheck(ctx context.Context) (errRet error) {
...
+		}()
+
+		var restartStatusResponse rdkweb.RestartStatusResponse
+		err = json.NewDecoder(resp.Body).Decode(&restartStatusResponse)
+		if err != nil {
+			// Only debug-log errors from JSON decoding. Interacting with older
+			// viam-server instances will result in a decoding error, as empty HTML
+			// will be returned.
+			s.logger.Debugf("ignoring error from restart allowed check for %s: %v",
+				SubsysName, err.Error())
```
Otterverse, line 334: Burying errors doesn't feel right. If we ever broke this on EITHER side, end users wouldn't see any indication why.

I'd much rather see you check the http response code and explicitly consider a non 2XX response (in this case, it should be a 400) to mean an older version of viam-server is in use. Only proceed to decoding the body if the response was good.

Otterverse (changes requested):
Looks good to me for the most part. Couple small things, but the main one is checking the http response status instead of waiting for json decode to fail.

subsystems/viamserver/viamserver.go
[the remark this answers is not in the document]
benjirewis: Removed named returns.

subsystems/viamserver/viamserver.go
benjirewis: Good point; done.

subsystems/viamserver/viamserver.go
benjirewis: Great point; done.

subsystems/viamserver/viamserver.go
benjirewis: > That said, if you feel it's not "dangerous" (it will be a signaled/controlled shutdown via SIGTERM after all) to hit those false positives occasionally, then that's fine by me. Simpler is nice in a lot of ways.

Yep that's the general sentiment! Will stick with the simpler method since it's alright with you.

subsystems/viamserver/viamserver.go
benjirewis: Done.

subsystems/viamserver/viamserver.go
benjirewis: Updated log to be `viam-server updated from [previous version] to 0.50.0`.

subsystems/viamserver/viamserver.go
```diff
@@ -294,17 +297,84 @@ func (s *viamServer) HealthCheck(ctx context.Context) (errRet error) {
...
+		//nolint:gosec
+		client := &http.Client{Transport: &http.Transport{TLSClientConfig: &tls.Config{InsecureSkipVerify: true}}}
+
+		resp, err := client.Do(req)
+		if err != nil {
+			return false, errw.Wrapf(err, "checking whether %s allows restart", SubsysName)
+		}
+
+		defer func() {
+			utils.UncheckedError(resp.Body.Close())
```
dgottlieb, line 326: I think Go has addressed the cases of defering lambdas within a for-loop capturing loop-scoped variables?

dgottlieb (approved):

Otterverse (approved):
LGTM (as soon as RDK updates and you can remove the go replace, of course)

benjirewis:
`rdk` v0.51.0 should be released tomorrow; will update then.

go.mod
```diff
@@ -64,32 +128,53 @@ require (
...
+	google.golang.org/genproto v0.0.0-20240903143218-8af14fe29dc1 // indirect
 	google.golang.org/genproto/googleapis/api v0.0.0-20240827150818-7e3bb234dfed // indirect
 	google.golang.org/genproto/googleapis/rpc v0.0.0-20240903143218-8af14fe29dc1 // indirect
 	gopkg.in/square/go-jose.v2 v2.6.0 // indirect
+	gopkg.in/src-d/go-billy.v4 v4.3.2 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 	nhooyr.io/websocket v1.8.9 // indirect
 )
+
+replace go.viam.com/rdk => ../rdk
```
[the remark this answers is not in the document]
benjirewis: Done.

PR #43, opened 2024-11-19
[APP-6881] Add disable wifi power save option

ale7714:


subsystems/provisioning/provisioning.go
```diff
@@ -359,3 +371,30 @@ func (w *Provisioning) updateHotspotSSID(cfg *Config) {
...
+
+	isNew, err := agent.WriteFileIfNew(wifiPowerSaveFilepath, []byte(contents))
+	if err != nil {
+		return errw.Wrap(err, "error writing wifi-powersave.conf")
+	}
+
+	if isNew {
+		w.logger.Infof("Updated %s to: %q", wifiPowerSaveFilepath, contents)
+		// Reload NetworkManager to apply changes
+		if err := w.nm.Reload(0); err != nil {
```
ale7714, line 394: I confirmed looking at network manager logs that the config file logs without warning or errors

ale7714:
@Otterverse I can only get the change to apply by actually rebooting the device. Rebooting NetworkManager is not enough.

Otterverse (approved):
LGTM! I think we can get a beta out tomorrow and start some testing.

PR #44, opened 2024-11-19
[APP-7026] Handle rfkill, no-wifi-device cases

Otterverse:
This fixes two cases... the first, in the bug, is when the wifi is blocked by rfkill (aka, "airplane mode".) This has recently (past week) become the default for new Raspberry Pi image installs unless the user specifically adds a wifi network during the image creation process.

The second case is for devices that have no wifi adapters at all. Previously, results were undefined, and provisioning would continue trying to execute, scan, start hotspots, etc. even if that wasn't possible. Now the lack of wifi hardware causes provisioning to self-disable until the next restart of agent.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,7 +4,7 @@ go 1.23.1
 
 require (
 	github.com/Masterminds/semver/v3 v3.3.0
-	github.com/Otterverse/gonetworkmanager/v2 v2.2.0
+	github.com/Otterverse/gonetworkmanager/v2 v2.2.1
 	github.com/google/uuid v1.6.0
 	github.com/jessevdk/go-flags v1.6.1
 	github.com/nightlyone/lockfile v1.0.0
--- a/subsystems/provisioning/provisioning.go
+++ b/subsystems/provisioning/provisioning.go
@@ -29,7 +29,6 @@ type Provisioning struct {
 	monitorWorkers sync.WaitGroup
 
 	// blocks start/stop/etc operations
-	// holders of this lock must use HealthySleep to respond to HealthChecks from the parent agent during long operations
 	opMu     sync.Mutex
 	running  bool
 	disabled bool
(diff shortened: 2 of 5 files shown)


go.mod
```diff
@@ -4,7 +4,7 @@ go 1.23.1
 
 require (
 	github.com/Masterminds/semver/v3 v3.3.0
-	github.com/Otterverse/gonetworkmanager/v2 v2.2.0
+	github.com/Otterverse/gonetworkmanager/v2 v2.2.1
```
Otterverse, line 7: Small patch over there to add GetPropertyRadioFlags()

subsystems/provisioning/setup.go
```diff
@@ -139,3 +141,24 @@ func (w *Provisioning) initDevices() error {
...
 }
+
+func (w *Provisioning) enableWifi(ctx context.Context) error {
+	if err := w.nm.SetPropertyWirelessEnabled(true); err != nil {
+		return err
+	}
+
+	timeoutCtx, cancel := context.WithTimeout(ctx, time.Second*10)
+	defer cancel()
+	for {
```
Otterverse, line 152: Without this loop the first scan fails and throws an error, as it takes a couple seconds for the radio to power on.

utils.go
```diff
@@ -332,12 +332,3 @@ func ConvertAttributes[T any](attributes *structpb.Struct) (*T, error) {
 
 	return newConfig, nil
 }
-
-func Sleep(ctx context.Context, timeout time.Duration) bool {
```
Otterverse, line 336: This is just drive-by cleanup. Nothing uses this anymore.

subsystems/provisioning/provisioning.go
```diff
@@ -124,6 +124,18 @@ func (w *Provisioning) getNM() (gnm.NetworkManager, error) {
 		return nil, nmErr
 	}
 
+	flags, err := nm.GetPropertyRadioFlags()
+	if err != nil {
```
ale7714, line 128: 👍

ale7714 (approved):

PR #45, opened 2024-11-20
Reactivate connection on powersave change, misc cleanup.

Otterverse:
Doing a PR to Ale's branch for discussion. Needs to merge before that (#43) merges.

Tested on my Pi 5 and it works as expected. Only "catch" now is that if powersave is set explicitly, then the setting is removed entirely, the last remaining state "remains" until the next reboot (as the driver itself initially turned on powersave during boot.) That is, assuming the user hasn't manually configured their own setting.

### DIFF
--- a/subsystems/provisioning/definitions.go
+++ b/subsystems/provisioning/definitions.go
@@ -29,6 +29,10 @@ const (
 	ConnCheckFilepath = "/etc/NetworkManager/conf.d/80-viam.conf"
 	ConnCheckContents = "[connectivity]\nuri=http://packages.viam.com/check_network_status.txt\ninterval=300\n"
 
+	wifiPowerSaveFilepath        = "/etc/NetworkManager/conf.d/81-viam-wifi-powersave.conf"
+	wifiPowerSaveContentsDefault = "# This file intentionally left blank.\n"
+	wifiPowerSaveContentsDisable = "[connection]\n# Explicitly disable\nwifi.powersave = 2\n"
+	wifiPowerSaveContentsEnable  = "[connection]\n# Explicitly enable\nwifi.powersave = 3\n"
 	NetworkTypeWifi    = "wifi"
 	NetworkTypeWired   = "wired"
 	NetworkTypeHotspot = "hotspot"
@@ -366,8 +370,8 @@ type Config struct {
 	// Computed from HotspotPrefix and Manufacturer
 	hotspotSSID string
 
-	// When true, it will disable power save for all wifi connections managed by NetworkManager.
-	DisableWifiPowerSave *bool `json:"disable_wifi_power_save"`
+	// If set, will explicitly enable or disable power save for all wifi connections managed by NetworkManager.
+	WifiPowerSave *bool `json:"wifi_power_save"`
 }
 
 // Timeout allows parsing golang-style durations (1h20m30s) OR seconds-as-float from/to json.
--- a/subsystems/provisioning/grpc.go
+++ b/subsystems/provisioning/grpc.go
@@ -16,7 +16,7 @@ func (w *Provisioning) startGRPC() error {
 	bind := PortalBindAddr + ":4772"
 	lis, err := net.Listen("tcp", bind)
 	if err != nil {
-		return errw.Wrapf(err, "error listening on: %s", bind)
+		return errw.Wrapf(err, "listening on: %s", bind)
 	}
 
 	w.grpcServer = grpc.NewServer(grpc.WaitForHandlers(true))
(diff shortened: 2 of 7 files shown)


subsystems/provisioning/definitions.go
```diff
@@ -29,6 +29,10 @@ const (
 	ConnCheckFilepath = "/etc/NetworkManager/conf.d/80-viam.conf"
 	ConnCheckContents = "[connectivity]\nuri=http://packages.viam.com/check_network_status.txt\ninterval=300\n"
 
+	wifiPowerSaveFilepath        = "/etc/NetworkManager/conf.d/81-viam-wifi-powersave.conf"
```
Otterverse, line 32: These belonged here with similar definitions for the other files.

Also, changed filename to avoid potential collisions with user-managed/written files.

subsystems/provisioning/definitions.go
```diff
@@ -29,6 +29,10 @@ const (
...
+	wifiPowerSaveContentsDefault = "# This file intentionally left blank.\n"
```
Otterverse, line 33: If not set, we basically want to not set anything. Explicitly setting 0 or 1 might still override another file in conf.d that a user put in place.

subsystems/provisioning/definitions.go
```diff
@@ -366,8 +370,8 @@ type Config struct {
 	// Computed from HotspotPrefix and Manufacturer
 	hotspotSSID string
 
-	// When true, it will disable power save for all wifi connections managed by NetworkManager.
-	DisableWifiPowerSave *bool `json:"disable_wifi_power_save"`
+	// If set, will explicitly enable or disable power save for all wifi connections managed by NetworkManager.
+	WifiPowerSave *bool `json:"wifi_power_save"`
```
Otterverse, line 374: Double negatives are... weird. Changed to a postive. True == on == powersave. Let me know if that doesn't sound okay though.

subsystems/provisioning/grpc.go
```diff
@@ -16,7 +16,7 @@ func (w *Provisioning) startGRPC() error {
 	bind := PortalBindAddr + ":4772"
 	lis, err := net.Listen("tcp", bind)
 	if err != nil {
-		return errw.Wrapf(err, "error listening on: %s", bind)
+		return errw.Wrapf(err, "listening on: %s", bind)
```
Otterverse, line 19: A bunch of these scattered around. Errors no need to say the word "error" as it's implicit in the fact that it IS an error. Saw one in the new code, and cleaned up all my own uses of this elsewhere to match.

subsystems/provisioning/networkmanager.go
```diff
@@ -175,13 +175,13 @@ func (w *Provisioning) StartProvisioning(ctx context.Context, inputChan chan<- u
...
 	if err := w.activateConnection(ctx, w.Config().HotspotInterface, w.Config().hotspotSSID); err != nil {
-		return errw.Wrap(err, "error starting provisioning mode hotspot")
+		return errw.Wrap(err, "starting provisioning mode hotspot")
 	}
 
 	// start portal with ssid list and known connections
 	if err := w.startPortal(inputChan); err != nil {
 		err = errors.Join(err, w.deactivateConnection(w.Config().HotspotInterface, w.Config().hotspotSSID))
-		return errw.Wrap(err, "could not start web/grpc portal")
+		return errw.Wrap(err, "starting web/grpc portal")
```
Otterverse, line 184: Another type of minor cleanup/standardization... Wrapped errors should provide context, and nothing more, as the error itself says what went wrong.

subsystems/provisioning/definitions.go
ale7714: good call

subsystems/provisioning/provisioning.go
```diff
@@ -372,27 +364,34 @@ func (w *Provisioning) updateHotspotSSID(cfg *Config) {
...
 	if isNew {
 		w.logger.Infof("Updated %s to: %q", wifiPowerSaveFilepath, contents)
 		// Reload NetworkManager to apply changes
 		if err := w.nm.Reload(0); err != nil {
-			return errw.Wrap(err, "error reloading NetworkManager after wifi-powersave.conf update")
+			return errw.Wrap(err, "reloading NetworkManager after wifi-powersave.conf update")
+		}
+
+		ssid := w.netState.ActiveSSID(w.cfg.HotspotInterface)
+		if w.connState.getConnected() && ssid != "" {
```
ale7714, line 391: i guess we're checking with only do this if we're not in hotspot mode which is a good catch

ale7714 (approved):
thanks for the cleanup on the error logs too!

PR #49, opened 2024-12-04
Fix wifi candidate selection to include non-interface-bound candidates

Otterverse:
Wifi networks found from scan were associated with the scanning interface (e.g. wlan0) but externally configured wifi settings (e.g. prexisting networks) might not be associated with a specific interface. Thus, duplicate networks (one for "any" and one for "wlan0" were causing issues.

This PR should deduplicate the two maps. If a connection/settings has no interface associated (and it's wifi) it gets handled as though it WAS associated with the current HotspotInterface (can be configured, otherwise first detected wifi adapter.)

### DIFF
--- a/cmd/viam-agent/main.go
+++ b/cmd/viam-agent/main.go
@@ -183,7 +183,7 @@ func main() {
 	}
 
 	// wait until now when we (potentially) have a network logger to record this
-	globalLogger.Infof("Viam Agent Version: %s\nGit Revision: %s\n", agent.GetVersion(), agent.GetRevision())
+	globalLogger.Infof("Viam Agent Version: %s Git Revision: %s", agent.GetVersion(), agent.GetRevision())
 
 	// if FastStart is set, skip updates and start viam-server immediately, then proceed as normal
 	var fastSuccess bool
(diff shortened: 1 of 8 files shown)


subsystems/provisioning/networkmanager.go
```diff
@@ -607,11 +601,10 @@ func (w *Provisioning) mainLoop(ctx context.Context) {
...
 				cfg := NetworkConfig{
-					Type:      NetworkTypeWifi,
-					SSID:      userInput.SSID,
-					PSK:       userInput.PSK,
-					Priority:  priority,
-					Interface: w.Config().HotspotInterface,
+					Type:     NetworkTypeWifi,
+					SSID:     userInput.SSID,
+					PSK:      userInput.PSK,
+					Priority: priority,
```
ale7714, line 607: why is interface not needed?

ale7714: nvm i see bellow

ale7714 (approved):

ale7714 (approved):

cmd/viam-agent/main.go
```diff
@@ -183,7 +183,7 @@ func main() {
 	}
 
 	// wait until now when we (potentially) have a network logger to record this
-	globalLogger.Infof("Viam Agent Version: %s\nGit Revision: %s\n", agent.GetVersion(), agent.GetRevision())
+	globalLogger.Infof("Viam Agent Version: %s Git Revision: %s", agent.GetVersion(), agent.GetRevision())
```
Otterverse, line 186: drive-by... didn't realize I'd copy/pasted the newlines from the CLI "version" output here (where it makes no sense.)

PR #50, opened 2024-12-04
APP-6990 consume debug level from cloud config

abe-winter:
## What changed
- consume `debug` field from cloud config, instead of only from CLI switch
## Why
It's difficult to set this currently.
## Notes
Where is the right place for this? I put it in the attributes for the viam-agent subsystem, but the `map[string]DeviceSubsystemConfig` approach seems less ergonomic post unification.
## Manual test
With debug set to true:
> 2024-12-04T08:51:22.774Z	INFO	viam-agent	viamserver/viamserver.go:183	viam-server started
> 2024-12-04T08:51:22.774Z	**DEBUG**	viam-agent	agent/manager.go:277	starting background checks
> 2024-12-04T08:51:22.774Z	**DEBUG**	viam-agent	agent/manager.go:185	Checking cloud for update
> 2024-12-04T08:51:22.882Z	INFO	rdk.package_manager	packages/deferred_package_manager.go:143	cloud package manager created asyncronously

With debug set to false (second restart):
> 2024-12-04T08:51:28.572Z	INFO	viam-agent	viamserver/viamserver.go:183	viam-server started
> 2024-12-04T08:51:28.654Z	INFO	rdk.package_manager	packages/deferred_package_manager.go:143	cloud package manager created asyncronously

## Sample config
```json
{
  "agent": {
    "viam-agent": {
      "attributes": {"debug": true}
    }
}
```

### DIFF
--- a/manager.go
+++ b/manager.go
@@ -309,6 +309,7 @@ func (m *Manager) LoadSubsystems(ctx context.Context) error {
 	if err != nil {
 		m.logger.Error(errw.Wrap(err, "getting cached config"))
 	}
+	m.processConfig(cachedConfig)
 
 	for _, name := range registry.List() {
 		cfg, ok := cachedConfig[name]
@@ -422,6 +423,24 @@ func (m *Manager) dial(ctx context.Context) error {
 	return nil
 }
 
+// process non-subsystem effects of a config (i.e. agent-specific stuff that needs to happen when loading cache).
+func (m *Manager) processConfig(cfg map[string]*pb.DeviceSubsystemConfig) {
+	if agent, ok := cfg["viam-agent"]; ok {
+		if debug, ok := agent.Attributes.AsMap()["debug"].(bool); !ok {
+			m.logger.Error("viam-agent debug attribute is present but is not a bool")
+		} else {
+			// note: if this is present (true or false, rather than missing) it overrides the CLI debug switch.
+			// if the user removes the `debug` attribute, we don't revert to the CLI debug switch state. (we ideally should).
+			// note: this assumes m.logger is the global logger shared by the other subsystems.
+			if debug {
+				m.logger.SetLevel(logging.DEBUG)
+			} else {
+				m.logger.SetLevel(logging.INFO)
+			}
+		}
+	}
+}
+
 // GetConfig retrieves the configuration from the cloud, or returns a cached version if unable to communicate.
 func (m *Manager) GetConfig(ctx context.Context) (map[string]*pb.DeviceSubsystemConfig, time.Duration, error) {
 	if m.cloudConfig == nil {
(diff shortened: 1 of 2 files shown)


manager.go
```diff
@@ -422,6 +423,24 @@ func (m *Manager) dial(ctx context.Context) error {
 	return nil
 }
 
+// process non-subsystem effects of a config (i.e. agent-specific stuff that needs to happen when loading cache and when updating).
+func (m *Manager) processConfig(cfg map[string]*pb.DeviceSubsystemConfig) {
+	if agent, ok := cfg["viam-agent"]; ok {
+		if debug, ok := agent.GetAttributes().AsMap()["debug"].(bool); !ok {
```
ale7714, line 429: as an attribute within viam-agent makes sense to me

ale7714 (commented):
the logs around provisioning/network management will also be affected by this? 

I think so bc we have globalLogger but making sure

abe-winter:
> the logs around provisioning/network management will also be affected by this?

yup, they showed up in my test:

> 2024-12-04T08:51:20.968Z	DEBUG	viam-agent	**provisioning/provisioning.go:73**	Provisioning Config: &{Manufacturer:viam Model:custom FragmentID: HotspotInterface: HotspotPrefix:**** HotspotPassword:**** DisableDNSRedirect:false OfflineTimeout:120000000000 UserTimeout:300000000000 FallbackTimeout:600000000000 RoamingMode:false Networks:[] hotspotSSID: WifiPowerSave:<nil>}

abe-winter:
(unless something is copying globalLogger before it sets debug)

manager.go
```diff
@@ -422,6 +423,24 @@ func (m *Manager) dial(ctx context.Context) error {
 	return nil
 }
 
+// process non-subsystem effects of a config (i.e. agent-specific stuff that needs to happen when loading cache and when updating).
+func (m *Manager) processConfig(cfg map[string]*pb.DeviceSubsystemConfig) {
```
Otterverse, line 427: I feel like this doesn't really belong here, but in viamagent/viamagent.go file, as the Update() and NewSubsystem() functions there are specifically meant to handle config changes. (Though I get there's the subtle problem of how early in startup can this be set.)

I suppose this is fine for now though. There's a scope coming down the pipe likely to reorganize the config structure soon, so we can consolidate configuration handling better after/during that I think. Viam-server's "fast_start" and "start_timeout" are other special cases that are handled entirely uniquely too right now.

manager.go
```diff
@@ -422,6 +423,24 @@ func (m *Manager) dial(ctx context.Context) error {
...
+	if agent, ok := cfg["viam-agent"]; ok {
+		if debug, ok := agent.GetAttributes().AsMap()["debug"].(bool); !ok {
+			m.logger.Error("viam-agent debug attribute is present but is not a bool")
```
Otterverse, line 430: nit: Not a correct error message. This returns false (!ok) if it's absent OR not a bool. You can cast an empty interface to bool essentially.

Otterverse (approved):
One small nit, but otherwise looks fine!

manager.go
abe-winter: yeah I modeled after fast_start -- agree with changing (but later)

manager.go
abe-winter: oops, good catch, fixed in 8ba8315

PR #51, opened 2024-12-05
[APP-6886] Add option to reboot device after a certain amount of time

ale7714:


subsystems/provisioning/definitions.go
```diff
@@ -370,6 +376,10 @@ type Config struct {
 
 	// If set, will explicitly enable or disable power save for all wifi connections managed by NetworkManager.
 	WifiPowerSave *bool `json:"wifi_power_save"`
+
+	// If set, will reboot the device after it has been offline for this duration
+	// 0 will disable this feature.
+	DeviceRebootAfterOfflineMinutes int `json:"device_reboot_after_offline_minutes"`
```
ale7714, line 382: I consider doing `device_reboot_after_offline_hours` while not super granular. I also can't imagine you would want less than one hour

Otterverse, line 382: Would prefer to see this be a `Timeout` instead of `int` as then the json gets parsed automatically for timeouts.

subsystems/provisioning/definitions.go
```diff
@@ -297,6 +297,12 @@ func ConfigFromJSON(defaultConf Config, jsonBytes []byte) (*Config, error) {
 		return &conf, errw.Errorf("timeout values cannot be less than %s", time.Duration(minTimeout))
 	}
 
+	if conf.DeviceRebootAfterOfflineMinutes != 0 &&
+		conf.DeviceRebootAfterOfflineMinutes < int(time.Duration(conf.OfflineTimeout).Minutes()) &&
+		conf.DeviceRebootAfterOfflineMinutes < int(time.Duration(conf.UserTimeout).Minutes()) {
+		return &conf, errw.Errorf("device_reboot_after_offline_minutes cannot be less than offline_timeout or user_timeout")
+	}
```
Otterverse, line 304: Would prefer to see native time parsing, not converting stuff to ints. Look at the other timeout(s) above. If the type here is Timeout, all the parsing happens automagically, so no need to cast and stuff here. It can just follow the pattern of the other timeouts (directly above.)

subsystems/provisioning/definitions.go
Otterverse: We may need to change the default parsing to "minutes" for the timeout values (currently it assumes bare numbers are seconds.)

subsystems/provisioning/networkmanager.go
```diff
@@ -691,6 +692,16 @@ func (w *Provisioning) mainLoop(ctx context.Context) {
 
 		if pMode {
 			// complex logic, so wasting some variables for readability
+			deviceRebootAfterOfflineDuration := time.Duration(w.cfg.DeviceRebootAfterOfflineMinutes) * time.Minute
+
+			if w.cfg.DeviceRebootAfterOfflineMinutes > 0 && now.After(lastOnline.Add(deviceRebootAfterOfflineDuration)) {
+				w.logger.Infof("device has been offline for more than %s minutes, rebooting", deviceRebootAfterOfflineDuration)
+				syscall.Sync()
+				err := syscall.Reboot(syscall.LINUX_REBOOT_CMD_RESTART)
```
Otterverse, line 700: This feels very dangerous. This feels like it'd be a rugpull to running processes. Equivalent to `SYSREQ+b` Does it actually do a normal shutdown? If not, maybe exec "systemctl reboot" to guarantee a normal restart.

subsystems/provisioning/definitions.go
Otterverse: Minutes makes the most sense. All the timeouts should have the same "unit" and minutes feels correct. Also, with the way Timeout parsing works, you can use decimal units if needed. Or "1h"... take a look at the Timeout.Unmarshall() function. It treats bare numbers as seconds currently, but we can change that to minutes, but people can specify easier units if they want. "10h30m" and such (we don't have to document/advertise that extra feature!)

Otterverse (changes requested):
A few notes before I hit the road... No need to wait for me if you need to merge this before I get back on Tuesday though.

subsystems/provisioning/networkmanager.go
ale7714: I believe is the same according to the docs, underhood systemctl is running reboot but will update to use systemctl for consistency 

https://www.man7.org/linux/man-pages/man1/systemctl.1.html

Otterverse (changes requested):
More notes... sorry I didn't get all of these noted last time.

ale7714:
sorry removing review request because i kind of push some changes quickly but did not test or anything, and then got sidetracked

ale7714:
ready for review now

subsystems/provisioning/networkmanager.go
```diff
@@ -737,6 +739,20 @@ func (w *Provisioning) mainLoop(ctx context.Context) {
...
+		if offlineRebootTimeout {
+			w.logger.Infof("device has been offline for more than %s minutes, rebooting", w.cfg.DeviceRebootAfterOfflineMinutes)
+
+			syscall.Sync() // flush file system buffers
+			cmd := exec.Command("systemctl", "reboot")
+			output, err := cmd.CombinedOutput()
+			if err != nil {
+				w.logger.Error(errw.Wrapf(err, "running 'systemctl reboot' %s", output))
+			}
+			time.Sleep(time.Second * 100) // systemd DefaultTimeoutStopSec defaults to 90 seconds
```
Otterverse, line 753: Can't do this, or agent itself won't shut down when called and will just hang until the sleep is done. It will also fail healthchecks if the loop pauses here for too long. There's a function for pausing the loop though. See line 623 above for the previous use of the following:

```
if !w.mainLoopHealth.Sleep(ctx, time.Second*100) {
	return
}
w.logger.Errorf("Failed to reboot after %s time!", time.Second * 100)			
```

May also need a slightly longer timeout, as I BELIEVE that systemd 90 second default is per service, so if it's shutting down multiple services, it can definitely take longer than that. I'd think 2 or maybe even 5 minutes, since this isn't going to handle the failure case (where reboot doesn't happen.) Regardless, I'd DEFINITELY suggest logging an error if the timeout expires (included in suggestion above... rewrite as needed) so the user knows things are off the rails.

Otterverse (changes requested):
Getting there... couple more gotchas though.

subsystems/provisioning/networkmanager.go
ale7714: makes sense, thanks i missed there's a sleep on the main loop

Otterverse (approved):
Realized this was fixed, but didn't see a review request. LGTM anyway, but if you have more changes planned, go ahead.

PR #55, opened 2024-12-17
[APP-7257] Support older NM (>1.30.0) and hardware that cannot scan in hotspot mode

Otterverse:
EDIT: rc2 replaces rc1. Fix a timeout issue that caused erratic behavior on "normal" (read, previously fully working) systems.
EDIT: rc1 replaces rc0. Minor fix for unrelated reboot timeout code.

Two main changes...

1. If NM older than 1.38, avoid using RadioFlags to disable when wifi hardware is absent, and rely on later device scan.

2. Modify wifi scanning to detect repeated failures when in hotspot mode, and (after 3) disable such scans after warning the user. This means the network list in the web portal or provided to the mobile app may be several minutes out of date, as scan data had to come from before the hotspot was started. May want to advise users to reduce the fallback timeout to a lower interval if this causes them problems. Otherwise we hang out in hotspot for 10 minutes, even if a wifi network we know comes in range, we can't see it until hotspot exits to allow a scan.

Test build:

     ` "pin_url": "http://packages.viam.com/temp/viam-agent-v0.12.0-rc2-aarch64",`

### DIFF
--- a/subsystems/provisioning/definitions.go
+++ b/subsystems/provisioning/definitions.go
@@ -42,17 +42,18 @@ const (
 
 var (
 	DefaultConf = Config{
-		Manufacturer:       "viam",
-		Model:              "custom",
-		FragmentID:         "",
-		HotspotPrefix:      "viam-setup",
-		HotspotPassword:    "viamsetup",
-		DisableDNSRedirect: false,
-		RoamingMode:        false,
-		OfflineTimeout:     Timeout(time.Minute * 2),
-		UserTimeout:        Timeout(time.Minute * 5),
-		FallbackTimeout:    Timeout(time.Minute * 10),
-		Networks:           []NetworkConfig{},
+		Manufacturer:                    "viam",
+		Model:                           "custom",
+		FragmentID:                      "",
+		HotspotPrefix:                   "viam-setup",
+		HotspotPassword:                 "viamsetup",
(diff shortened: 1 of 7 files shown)


ale7714 (approved):
Thanks for fixing the bug. I haven't tested it but code looks good.

subsystems/provisioning/definitions.go
```diff
@@ -298,9 +300,11 @@ func ConfigFromJSON(defaultConf Config, jsonBytes []byte) (*Config, error) {
 	}
 
 	if conf.DeviceRebootAfterOfflineMinutes != 0 &&
-		conf.DeviceRebootAfterOfflineMinutes < conf.OfflineTimeout ||
-		conf.DeviceRebootAfterOfflineMinutes < conf.UserTimeout {
-		return &conf, errw.Errorf("device_reboot_after_offline_minutes cannot be less than offline_timeout or user_timeout")
+		(conf.DeviceRebootAfterOfflineMinutes < conf.OfflineTimeout || conf.DeviceRebootAfterOfflineMinutes < conf.UserTimeout) {
```
Otterverse, line 303: It took me WAY too long to figure out this was the issue. One pair of missing parenthesis! So easy to miss!

ale7714 (approved):

PR #56, opened 2025-01-09
[APP-7364] [APP-7366] [RSDK-9684] [APP-7154] Add more logging, reduce monitoring loop time, misc small fixes.

Otterverse:
Quick PR for easy fixes.

Needs testing for the linked Jira issues

1. Install
2. Upgrade/downgrade viam-agent and ensure a "disabled" systemd service remains disabled after upgrade.
3. Confirm captive portal works for new wifi (mainly that browser doesn't throw errors) and fixes capitalization issue.
4. Corrupt /etc/viam.json, ensure errors are clear and file is renamed
5. Make sure it runs okay for a while (15+ minutes) with the new, shorter timeout.
6. Kill viam-server (or otherwise make it fail) and confirm it logs and restarts quickly.

### DIFF
--- a/manager.go
+++ b/manager.go
@@ -24,7 +24,7 @@ import (
 )
 
 const (
-	minimalCheckInterval  = time.Second * 60
+	minimalCheckInterval  = time.Second * 5
 	defaultNetworkTimeout = time.Second * 15
 	// stopAllTimeout must be lower than systemd subsystems/viamagent/viam-agent.service timeout of 4mins
 	// and higher than subsystems/viamserver/viamserver.go timeout of 2mins.
--- a/subsystems/provisioning/templates/index.html
+++ b/subsystems/provisioning/templates/index.html
@@ -37,7 +37,7 @@
     <div class="form-group">
       <label for="network">Network</label>
       {{if eq (len .VisibleSSIDs) 0}}
-        <input type="text" name="ssid" placeholder="Enter Wifi SSID" id="network" required>
+        <input type="text" name="ssid" placeholder="Enter Wifi SSID" id="network" required autocorrect="off" autocapitalize="off">
       {{else}}
         <div class="select-border">
           <select name="ssid" id="network" required>
@@ -52,14 +52,14 @@
 
     <div class="form-group">
       <label for="password">Password</label>
-      <input type="text" name="password" id="password" placeholder="Password">
+      <input type="text" name="password" id="password" placeholder="Password" autocorrect="off" autocapitalize="off">
     </div>
     {{end}}
 
     {{if not .IsConfigured}}
     <div class="form-group">
       <label for="viamconfig">Device Config</label>
-      <textarea type="textarea" name="viamconfig" id="viamconfig" required placeholder="No config found on device. Paste your viam.json file here."></textarea>
+      <textarea type="textarea" name="viamconfig" id="viamconfig" required placeholder="No config found on device. Paste your viam.json file here." autocorrect="off" autocapitalize="off"></textarea>
     </div>
     {{end}}
 
--- a/subsystems/viamagent/viamagent.go
+++ b/subsystems/viamagent/viamagent.go
@@ -123,14 +123,14 @@ func Install(logger logging.Logger) error {
 		return errw.Wrap(err, "getting service file path")
 	}
 
-	//nolint:gosec
-	if err := os.MkdirAll(filepath.Dir(serviceFilePath), 0o755); err != nil {
-		return errw.Wrapf(err, "creating directory %s", filepath.Dir(serviceFilePath))
-	}
+	// use this later to avoid re-enabling an existing agent service a user might have disabled
+	_, err = os.Stat(serviceFilePath)
+	newInstall := err != nil
 
 	logger.Infof("writing systemd service file to %s", serviceFilePath)
-	//nolint:gosec
-	if err := os.WriteFile(serviceFilePath, serviceFileContents, 0o644); err != nil {
+
+	newFile, err := agent.WriteFileIfNew(serviceFilePath, serviceFileContents)
+	if err != nil {
 		return errw.Wrapf(err, "writing systemd service file %s", serviceFilePath)
 	}
 
@@ -143,17 +143,21 @@ func Install(logger logging.Logger) error {
 		}
 	}
 
-	logger.Infof("enabling systemd viam-agent service")
-	cmd = exec.Command("systemctl", "daemon-reload")
-	output, err = cmd.CombinedOutput()
-	if err != nil {
-		return errw.Wrapf(err, "running 'systemctl daemon-reload' output: %s", output)
+	if newFile {
+		cmd = exec.Command("systemctl", "daemon-reload")
+		output, err = cmd.CombinedOutput()
+		if err != nil {
+			return errw.Wrapf(err, "running 'systemctl daemon-reload' output: %s", output)
+		}
 	}
 
-	cmd = exec.Command("systemctl", "enable", "viam-agent")
-	output, err = cmd.CombinedOutput()
-	if err != nil {
-		return errw.Wrapf(err, "running 'systemctl enable viam-agent' output: %s", output)
+	if newInstall {
+		logger.Infof("enabling systemd viam-agent service")
+		cmd = exec.Command("systemctl", "enable", "viam-agent")
+		output, err = cmd.CombinedOutput()
+		if err != nil {
+			return errw.Wrapf(err, "running 'systemctl enable viam-agent' output: %s", output)
+		}
 	}
 
 	_, err = os.Stat("/etc/viam.json")
--- a/cmd/viam-agent/main.go
+++ b/cmd/viam-agent/main.go
@@ -140,6 +140,8 @@ func main() {
 		// If the local /etc/viam.json config is corrupted, invalid, or missing (due to a new install), we can get stuck here.
 		// Rename the file (if it exists) and wait to provision a new one.
 		if !errors.Is(err, fs.ErrNotExist) {
+			globalLogger.Error(errors.Wrapf(err, "reading %s", absConfigPath))
+			globalLogger.Warn("renaming %s to %s.old", absConfigPath, absConfigPath)
 			if err := os.Rename(absConfigPath, absConfigPath+".old"); err != nil {
 				// if we can't rename the file, we're up a creek, and it's fatal
 				globalLogger.Error(errors.Wrapf(err, "removing invalid config file %s", absConfigPath))
--- a/subsystems/viamserver/viamserver.go
+++ b/subsystems/viamserver/viamserver.go
@@ -172,6 +172,9 @@ func (s *viamServer) Start(ctx context.Context) error {
 				s.logger.Errorw("non-zero exit code", "exit code", s.lastExit)
 			}
 		}
+		if s.shouldRun {
+			s.logger.Infof("%s exited unexpectedly and will be restarted shortly", SubsysName)
+		}
 		close(s.exitChan)
 	}()
 


subsystems/provisioning/templates/index.html
```diff
@@ -37,7 +37,7 @@ <h2>Smart Machine Setup</h2>
     <div class="form-group">
       <label for="network">Network</label>
       {{if eq (len .VisibleSSIDs) 0}}
-        <input type="text" name="ssid" placeholder="Enter Wifi SSID" id="network" required>
+        <input type="text" name="ssid" placeholder="Enter Wifi SSID" id="network" required autocorrect="off" autocapitalize="off">
```
maxhorowitz, line 40: For context these are "on" if omitted and that was not allowing you to connect with a SSID that was lowercased? Nice find if so 😄

subsystems/viamagent/viamagent.go
```diff
@@ -123,14 +123,14 @@ func Install(logger logging.Logger) error {
 		return errw.Wrap(err, "getting service file path")
 	}
 
-	//nolint:gosec
-	if err := os.MkdirAll(filepath.Dir(serviceFilePath), 0o755); err != nil {
-		return errw.Wrapf(err, "creating directory %s", filepath.Dir(serviceFilePath))
-	}
+	// use this later to avoid re-enabling an existing agent service a user might have disabled
```
maxhorowitz, line 126: Want to confirm my understanding...there are users who have Viam agent installed but who don't want  it to automatically start up on boot, so they've "disabled" auto start. When they upgrade their Viam agent, the configuration file where they've "disabled" the auto start gets overwritten such that startup on boot becomes true...and that's what you're fixing in this PR.

Copied the following snippet from the ticket - is this the final product (i.e. the user now has to manually restart Viam agent after an upgrade)?
```
Jan 07 16:11:09 pluto viam-agent[911]: 2025-01-07T22:11:09.496Z        INFO        viam-agent        agent/subsystem.go:355        viam-agent updated from 0.11.0 to 0.12.0
Jan 07 16:11:10 pluto viam-agent[911]: 2025-01-07T22:11:09.515Z        INFO        viam-agent        viamagent/viamagent.go:131        writing systemd service file to /usr/local/lib/systemd/system/viam-agent.service
Jan 07 16:11:10 pluto viam-agent[911]: 2025-01-07T22:11:09.515Z        INFO        viam-agent        viamagent/viamagent.go:146        enabling systemd viam-agent service
Jan 07 16:11:10 pluto viam-agent[911]: 2025-01-07T22:11:10.091Z        INFO        viam-agent        viamagent/viamagent.go:168        Install complete. Please (re)start the service with 'systemctl restart viam-agent' when ready.
```

manager.go
```diff
@@ -24,7 +24,7 @@ import (
 )
 
 const (
-	minimalCheckInterval  = time.Second * 60
+	minimalCheckInterval  = time.Second * 5
```
maxhorowitz, line 27: What's the monitoring loop checking for on a super high level? Is this monitoring for FTDC data in the viam-server? What other things are "monitored" on a regular basis via the agent ?

subsystems/viamagent/viamagent.go
```diff
@@ -123,14 +123,14 @@ func Install(logger logging.Logger) error {
...
+	_, err = os.Stat(serviceFilePath)
+	newInstall := err != nil
```
maxhorowitz, line 128: Worth checking for a file does not exist error?

maxhorowitz (approved):
Code scan ✅ 
Have some questions which are mainly aimed at me getting up to speed! I will manually test these fixes later today.

manager.go
Otterverse: This is how often it fetches a new config from the cloud. The cloud itself can say "check again after X time") but that's currently not implemented, so it falls to this.

The main loop basically: check for new config > apply changes/updates (if there) > start subsystems that should be started > check health of subsystems > repeat

So the check interval determins (roughly) the overall timing. At 5 seconds, it may have trouble keeping up if there are slow responses or other work to do, but that should be fine.

subsystems/provisioning/templates/index.html
Otterverse: Micheal Lee found/suggested this fix. But yeah, I guess some phones/browsers see an empty text field, and if you type a password that's a normal "word" it auto-capitalizes it, which isn't ideal for passwords.

We don't use a "password" type field here, because I wanted this to be visible before submitting. Otherwise it's way too easy to accidentally typo and then it takes several minutes for the device to try out the thing, fail, and restart a hotspot.

subsystems/viamagent/viamagent.go
Otterverse: So if someone installed viam-agent, then runs `systemctl disable viam-agent` it won't start at boot. The problem is that, previously, when we did an upgrade, this same code got called, and it ALWAYS enabled the service again. Now it _should_ only enable it on first/new install.

The snippet is the old behavior, where it shows "enabling systemd viam-agent service" was the problem. That should only happen on new installs now.

subsystems/viamagent/viamagent.go
Otterverse: This covers edge cases like file corruption too. If we can't stat, then it's best to treat it as a new install, IMHO. But a good thought! In many other cases it'd be worth differentiating.

manager.go
maxhorowitz: Thanks

subsystems/viamagent/viamagent.go
maxhorowitz: Working on getting manual testing setup. Lmk if I'm moving in right direction. Using an RPi5 because agent runs on linux (and I use Mac). Should I be:

1. pulling a config from prod (without actually starting the viam-server)
2. check out your branch and run `make`
3. run the binary that I've created which includes the changes in your PR
4. _validate the viam-agent starts at reboot of the RPi5_
5. run `systemctl disable viam-agent`
6. **upgrade the `viam-agent` on the RPi5 by doing what exactly?**
7. _validate that the `viam-agent` service isn't running on 2nd reboot_

Please let me know gaps here. Thanks

maxhorowitz: Validated by disabling the agent and upgrading from a stable version
```json
{
  "agent": {
    "viam-agent": {
      "pin_version": "0.12.0"
    },
    "viam-server": {
      "release_channel": "stable",
      "attributes": {}
    }
  }
}
```
to a pinned URL:
```json
{
  "agent": {
    "viam-agent": {
      "pin_url": "file:///home/jeep/dev/agent/bin/viam-agent-custom-aarch64"
    },
    "viam-server": {
      "release_channel": "stable",
      "attributes": {}
    }
  }
}
```
After the upgrade and subsequent reboot, I validated the `viam-agent` system service wasn't running and was still disabled:

maxhorowitz: After re-enable and reboot the viam-agent starts up on its own 👍 
```
● viam-agent.service - Viam Services Agent
     Loaded: loaded (/usr/local/lib/systemd/system/viam-agent.service; enabled; preset: enabled)
     Active: active (running) since Fri 2025-01-10 13:56:50 EST; 1min 22s ago
   Main PID: 758 (viam-agent)
      Tasks: 10 (limit: 9247)
        CPU: 4.953s
     CGroup: /system.slice/viam-agent.service
             └─758 /opt/viam/bin/viam-agent --config /etc/viam.json
```

maxhorowitz:
@Otterverse - working through the cases you need tested - I'm not familiar with the wifi setup flow - how should I be testing
> Confirm captive portal works for new wifi (mainly that browser doesn't throw errors) and fixes capitalization issue.
And does the following mean adding an unknown field to the cloud config? Otherwise, how would it be able to re-fetch its config if I change  one of its values? Or is it cached on the robot and startup?
> Corrupt /etc/viam.json, ensure errors are clear and file is renamed

Otterverse:
> @Otterverse - working through the cases you need tested - I'm not familiar with the wifi setup flow - how should I be testing
> 
> > Confirm captive portal works for new wifi (mainly that browser doesn't throw errors) and fixes capitalization issue.
> 

After installing and using the custom version, you can delete the wifi network with `nmcli` and it should go into provisioning mode. When it does, you can connect to the hotspot it makes, and verify that you can use the captive portal (you should be redirected automatically if you connect with your phone as a "sign in" page) to add a wifi network and it doesn't automatically capitalize the password.

> And does the following mean adding an unknown field to the cloud config? Otherwise, how would it be able to re-fetch its config if I change one of its values? Or is it cached on the robot at startup?
> 
> > Corrupt /etc/viam.json, ensure errors are clear and file is renamed

You can always get the config from the link in App for the machine. Click the status bar "offline" notification to make it drop down and give details. There's a "machine credentials" button there which gives you the contents of /etc/viam.json. And to corrupt that file, I meant just make it invalid... open it up with a text editor like nano or whatever, and just delete some curly brackets or something to make it invalid json.

Ping me on slack if you have any questions. Thanks again for the work here!

maxhorowitz:
RE: https://github.com/viamrobotics/agent/pull/56#issuecomment-2587709242

## Findings

### (1) Does agent provisioning automatically start?
**Not sure that this is working as intended**. I was running both the agent (with the pinned version pointing to my local agent which I built from your branch) and the server, and I used `nmcli con delete ****` to delete the active wlan connection - this didn't bring me directly into provisioning mode unless my process was wrong. In a separate tab, once I ran `sudo systemctl restart viam-agent` the agent-provisioning startup would happen and then I would be able to re-configure its wifi. _Note: I was testing this on an RPi5 that was plugged into a monitor (such that I could see both `viam-server` and `viam-agent` logs with/out internet connection so I'm pretty confident I'm not wrong_.

### (2) Does the captive portal work once the viam-agent is in provisioning mode?
Yes.

### (3) Does the lowercased wifi password get modified to uppercase?
No, as intended.

### (4) Does the `viam-server` start up properly after connecting to the Wifi network provided in the captive portal?
Not really. I'm seeing a continuous loop of restarting the `viam-server`. The error is `listen tcp  :8080: bind: address already in use`, and I've killed the individual process running `viam-server` on that port but I keep getting this same loop. I fixed by running `killall` on `viam-server`, which results in a successful restart of `viam-server` by the agent.

### (5) Once running, corrupt `/etc/viam.json` and ensure the file is overwritten to clean version.
This works.

maxhorowitz:
Update about the following observation on Thursday:

> ### (1) Does agent provisioning automatically start?
> Not sure that this is working as intended. I was running both the agent (with the pinned version pointing to my local agent which I built from your branch) and the server, and I used nmcli con delete **** to delete the active wlan connection - this didn't bring me directly into provisioning mode unless my process was wrong. In a separate tab, once I ran sudo systemctl restart viam-agent the agent-provisioning startup would happen and then I would be able to re-configure its wifi. Note: I was testing this on an RPi5 that was plugged into a monitor (such that I could see both viam-server and viam-agent logs with/out internet connection so I am pretty confident about my observations being true - whether or not my process was wrong is a separate story).

Confirmed I had separate installs of viam-server and viam-agent which was the culprit. The gap in my knowledge was _I thought `viam-server` system service would be installed and run by the existing `viam-agent` system service_, but I now see that there's a "CGroup" within the `viam-agent` system service that manages the two (agent and server) in tandem:
```
jeep@jeep:~ $ sudo systemctl status viam-agent | cat
● viam-agent.service - Viam Services Agent
     Loaded: loaded (/usr/local/lib/systemd/system/viam-agent.service; enabled; preset: enabled)
     Active: active (running) since Tue 2025-01-21 09:39:33 EST; 15min ago
   Main PID: 1983 (viam-agent)
      Tasks: 21 (limit: 9247)
        CPU: 21.591s
     CGroup: /system.slice/viam-agent.service
             ├─1983 /opt/viam/bin/viam-agent --config /etc/viam.json
             └─1996 /opt/viam/bin/viam-server -config /etc/viam.json
```
Is this the correct understanding?

If so, both (1) and (4) are working as expected (which is everything that I wasn't yet sure about) - thank you @Otterverse for coaching me through some of the testing 👍

Otterverse:
> Update about the following observation on Thursday:
> 
> > ### (1) Does agent provisioning automatically start?
> > Not sure that this is working as intended. I was running both the agent (with the pinned version pointing to my local agent which I built from your branch) and the server, and I used nmcli con delete **** to delete the active wlan connection - this didn't bring me directly into provisioning mode unless my process was wrong. In a separate tab, once I ran sudo systemctl restart viam-agent the agent-provisioning startup would happen and then I would be able to re-configure its wifi. Note: I was testing this on an RPi5 that was plugged into a monitor (such that I could see both viam-server and viam-agent logs with/out internet connection so I am pretty confident about my observations being true - whether or not my process was wrong is a separate story).
> 
> Confirmed I had separate installs of viam-server and viam-agent which was the culprit. The gap in my knowledge was _I thought `viam-server` system service would be installed and run by the existing `viam-agent` system service_, but I now see that there's a "CGroup" within the `viam-agent` system service that manages the two (agent and server) in tandem:
> 
> ```
> jeep@jeep:~ $ sudo systemctl status viam-agent | cat
> ● viam-agent.service - Viam Services Agent
>      Loaded: loaded (/usr/local/lib/systemd/system/viam-agent.service; enabled; preset: enabled)
>      Active: active (running) since Tue 2025-01-21 09:39:33 EST; 15min ago
>    Main PID: 1983 (viam-agent)
>       Tasks: 21 (limit: 9247)
>         CPU: 21.591s
>      CGroup: /system.slice/viam-agent.service
>              ├─1983 /opt/viam/bin/viam-agent --config /etc/viam.json
>              └─1996 /opt/viam/bin/viam-server -config /etc/viam.json
> ```
> 
> Is this the correct understanding?
> 
> If so, both (1) and (4) are working as expected (which is everything that I wasn't yet sure about) - thank you @Otterverse for coaching me through some of the testing 👍

Pretty much correct, but the fact that they're in a cgroup isn't "important" directly. Rather, just anything part of a systemd service will be part of the same cgroup by default. The way I'd phrase is that there's only one systemd service, which is for the agent. Viam server is a child process of the agent. That is, agent starts/stops/manages it directly (not through systemd.)

Sounds like we're good to go on merging this for now though. Thanks for the effort here!

PR #62, opened 2025-02-14
[APP-7635]: Define type for managing bluetooth WiFi provisioning.

maxhorowitz:
# Summary
Add type (without implementing a low-level Linux solution) that can be "plugged-in" to the Agent provisioning flow for spinning up BT, waiting for credentials, and cleanly shutting itself down.
# Included
1. [APP-7635]: Define type for bluetooth provisioning, which does not include any low-level bluetooth-stack specific assumptions about how we will read/write values to bluetooth characteristics on the peripheral.
2. I have _**NOT**_ added unit tests. I will do so once we can agree upon the public implementation added in this PR.

The following functionality is added in this PR:
```go
type bluetoothWiFiProvisioner struct {...}
func NewBluetoothWiFiProvisioner(...) (*bluetoothWiFiProvisioner, error)
func (bwp *BluetoothWiFiProvisioner) Start(ctx context.Context) error {...}
func (bwp *BluetoothWiFiProvisioner) Stop() error {...}
func (bwp *BluetoothWiFiProvisioner) RefreshAvailableNetworks(ctx context.Context, awns []*NetworkInfo) error {...}
func (bwp *BluetoothWiFiProvisioner) WaitForCredentials(ctx context.Context) (*userInput, error) {...}
```
# Example usage
The following code snippet describes how I plan to use the functionality added in this PR.
```go
func bluetoothWiFiProvisioningExample(ctx context.Context, logger logging.Logger, availableWiFiNetworksChan chan []*NetworkInfo) {
	ctx, cancel := context.WithCancel(ctx)
	defer cancel()

	bwp, err := ble.NewBluetoothWiFiProvisioner(ctx, logger, "Viam Agent") // STEP 1: Create instance that implements interface.
	if err != nil {
		logger.Fatal(err)
	}

	go func() {  // STEP 2: Run background job to update available WiFi networks.
		for {
			if ctx.Err() != nil {
				return
			}
			select {
			case <-ctx.Done():
				return
			case awn := <-availableWiFiNetworksChan:
				// 
				// Example []*NetworkInfo value (passed in via channel):
				// 
				// awn := []*NetworkInfo{
				// 		{
				// 			SSID:        "HomeWiFi",
				// 			Signal:    85,
				// 			Security: "WPA",
				// 		},
				// 		{
				// 			Ssid:        "GuestWiFi",
				// 			Signal:    50,
				// 			Security: "None",
				// 		},
				// 	},
				// }
				// 
				if err := bwp.RefreshAvailableNetworks(ctx, awn); err != nil {
					logger.Error(err)
					return
				}
			default:
				time.Sleep(time.Second * 3)
			}
		}
	}()

	if err := bwp.Start(ctx); err != nil { // STEP 3: Start advertising BLE to a client.
		logger.Fatal(err)
	}

	// Interface method naming implies this should be a blocking call.
	userInput, err := bwp.WaitForCredentials(ctx, true, true) // STEP 4: Wait for the client to send credentials over BT.
	if err != nil {
		logger.Fatal(err)
	}
	logger.Infof("User provided SSID: %s and Psk: %s, will attempt to connect with those WiFi credentials. "+
		"Once connected, will provision this device with the following cloud config: Robot Part Key ID: %s, "+
		"Robot Part Key: %s", userInput.SSID, userInput.PSK, userInput.PartD, userInput.Secret)

	if err := bwp.Stop(); err != nil { // STEP 5: Close BLE advertisement (tears down client connection if exists).
		logger.Fatal(err)
	}
}
```

[APP-7635]: https://viam.atlassian.net/browse/APP-7635?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[APP-7637]: https://viam.atlassian.net/browse/APP-7637?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


maxhorowitz:
Note to reviewers: once we can agree upon the interface(s) added here, I will add unit testing to make it lock tight.

subsystems/provisioning/bluetooth/bluetooth_service.go
```diff
@@ -0,0 +1,37 @@
...
+
+type bluetoothService interface {
+	startAdvertisingBLE(ctx context.Context) error
+	stopAdvertisingBLE() error
+	enableAutoAcceptPairRequest()
+
+	// Networks need to be written to a bluetooth service. Clients can then read this data.
+	writeAvailableNetworks(networks *AvailableWiFiNetworks) error
+
+	// Credentials that ae written by a client need to be extracted from a bluetooth service.
```
maxhorowitz, line 17:
```suggestion
	// Credentials that are written by a client need to be extracted from a bluetooth service.
```

subsystems/provisioning/bluetooth/bluetooth_service.go
```diff
@@ -0,0 +1,37 @@
...
+	"context"
+	"fmt"
+)
+
+type bluetoothService interface {
+	startAdvertisingBLE(ctx context.Context) error
+	stopAdvertisingBLE() error
+	enableAutoAcceptPairRequest()
+
+	// Networks need to be written to a bluetooth service. Clients can then read this data.
```
maxhorowitz, line 14:
```suggestion
	// Available WiFi networks need to be written to a bluetooth service. Clients read from this inputted data.
```

subsystems/provisioning/bluetooth/bluetooth_wifi_provisioner.go
```diff
@@ -0,0 +1,155 @@
...
+		RequiresPsk bool    `json:"requires_psk"`
+	} `json:"networks"`
+}
+
+func (awns *AvailableWiFiNetworks) ToBytes() ([]byte, error) {
+	return json.Marshal(awns)
+}
+
+// BluetoothWiFiProvisioner provides an interface for managing the bluetooth (bluetooth-low-energy) service as it pertains to WiFi setup.
+type BluetoothWiFiProvisioner interface {
```
maxhorowitz, line 39: I don't plan on needing to add anything to this file, it serves as a wrapper around the OS-specific bluetooth implementations, which are implemented behind the unexported `bluetoothService` interface. I know it is against Golang best practice to have unexported interfaces, and I'm far from an expert using generic types, so let me know if there are better ways to be designing this stuff.

subsystems/provisioning/bluetooth/bluetooth_service.go
```diff
@@ -0,0 +1,37 @@
+// Package ble contains an interface for interacting with the bluetooth stack on a Linux device, specifically with respect to provisioning.
+package ble
+
+import (
+	"context"
+	"fmt"
+)
+
+type bluetoothService interface {
```
maxhorowitz, line 9: I know it is against Golang best practice to have unexported interfaces, and I'm far from an expert using generic types, so let me know if there are better ways to be designing this feature. 

The purpose of this layer of abstraction was to protect the Linux-specific BT stuff from the core implementation of our `BluetoothWiFiProvisioner` interface. It isn't exported because I don't expect anyone to need it. If you have more specific questions, let me know, and I'll dive in.

subsystems/provisioning/bluetooth/bluetooth_wifi_provisioner.go
maxhorowitz, line 39: This is the exported interface which is to be consumed in the provisioning loop. In the interest of not "muddying" our provisioning loop flow, I keep this interface very simple. It only has things that you would need from the perspective of the provisioning loop, and it doesn't include anything about BT peripheral advertisement, characteristics, etc.

I get into the more "under-the-hood" functionality in the `bluetooth_service.go` file. There, I also have an interface so I can abstract away all Linux OS stuff.

subsystems/provisioning/bluetooth/bluetooth_service.go
benjirewis: It's tough for me to tell what's going on here without the implementer of this `bluetoothService` interface, but I'm definitely suspicious of the need for _any interface_ at all. 

Will you have multiple implementers of `bluetoothService`? Or multiple of `BluetoothWiFiProvisioner`? 

Using interfaces to have a "pimpl" (private implementation + public methods) design is definitely against Go best practices, but maybe I'm misunderstanding what you're doing here. Interfaces are generally helpful for _accepting_ multiple types of things that implement a common set of methods, not for hiding implementations. Apologies if you know these things already, but figured I'd come out strong and say I'm suspicious of the patterns you're using here wrt to the two interfaces (and one that's unexported.) 

I think my concern extends to your use of generics, as they seem to be how you're getting around your usage of interfaces.

subsystems/provisioning/bluetooth/bluetooth_wifi_provisioner.go
```diff
@@ -0,0 +1,155 @@
...
+	return nil
+}
+
+// Stop stops advertising a bluetooth service which (when enabled) accepts WiFi and Viam cloud config credentials.
+func (bwp *bluetoothWiFiProvisioner[T]) Stop() error {
+	return bwp.svc.stopAdvertisingBLE()
+}
+
+// Update updates the list of networks that are advertised via bluetooth as available.
+func (bwp *bluetoothWiFiProvisioner[T]) RefreshAvailableNetworks(ctx context.Context, awns *AvailableWiFiNetworks) error {
```
benjirewis, line 66: This method probably doesn't need a context if it's not passing it anywhere and is just immediately checking if it has errored.

subsystems/provisioning/bluetooth/bluetooth_wifi_provisioner.go
```diff
@@ -0,0 +1,155 @@
...
+		return nil, errors.New("should be waiting for either cloud credentials or WiFi credentials")
+	}
+	var ssid, psk, robotPartKeyID, robotPartKey string
+	var ssidErr, pskErr, robotPartKeyIDErr, robotPartKeyErr error
+	wg := &sync.WaitGroup{}
+	if requiresWiFiCredentials {
+		wg.Add(2)
+		utils.ManagedGo(
+			func() {
+				ssid, ssidErr = waitForBLEValue(ctx, bwp.svc.readSsid, "ssid")
```
benjirewis, line 85: If one of these four goroutines errors, do you want the other goroutines to stop?

subsystems/provisioning/bluetooth/bluetooth_wifi_provisioner.go
```diff
@@ -0,0 +1,155 @@
...
+		)
+	}
+	wg.Wait()
+	return &Credentials{
+		Ssid: ssid, Psk: psk, RobotPartKeyID: robotPartKeyID, RobotPartKey: robotPartKey,
+	}, multierr.Combine(ssidErr, pskErr, robotPartKeyIDErr, robotPartKeyErr)
+}
+
+// waitForBLE is used to check for the existence of a new value in a BLE characteristic.
+func waitForBLEValue(
```
benjirewis, line 118: [nit] This function isn't _really_ waiting for any BLE value, it's just calling a callback in a loop and continuing if there's a certain type of error. I wonder if the name should reflect that a bit better.

subsystems/provisioning/bluetooth/bluetooth_wifi_provisioner.go
```diff
@@ -0,0 +1,155 @@
...
+package ble
+
+import (
+	"context"
+	"encoding/json"
+	"runtime"
+	"sync"
+	"time"
+
+	"github.com/pkg/errors"
```
benjirewis, line 11: [nit] Not sure what agent's pattern is, but generally the standard library `errors` is a little more idiomatic.

subsystems/provisioning/bluetooth/bluetooth_wifi_provisioner.go
```diff
@@ -0,0 +1,155 @@
...
+		return v, nil
+	}
+}
+
+// NewBluetoothWiFiProvisioner returns a service which accepts credentials over bluetooth to provision a robot and its WiFi connection.
+func NewBluetoothWiFiProvisioner(ctx context.Context, logger logging.Logger, name string) (BluetoothWiFiProvisioner, error) {
+	switch os := runtime.GOOS; os {
+	case "linux":
+		fallthrough
+	case "windows":
```
benjirewis, line 148: Pls no.

benjirewis (commented):
Missing some of the picture here, but I thought I'd leave a round of comments before I'm OOO for most of next week. Feel free to ignore anything, too, more @Otterverse's + your code than mine.

subsystems/provisioning/bluetooth/bluetooth_service.go
maxhorowitz: Thanks for diving in --- and definitely appreciate explanations of pimpl and best practices because I'm always looking to level up in Golang. I've used pimpl interfaces in the past because I liked how it made testing modular, so that's what I was going for --- and happy to learn that's not how it should be done.

I added the `bluetoothService` interface with the idea that some day we could support more than just Linux. The implementation I had in mind was `linuxBluetoothService`, `windowsBluetoothService`, etc., so that we could abstract away OS-specific bluetooth differences from the business login inside of `BluetoothWiFiProvisioner`. That being said, **it is super unlikely** we get to that point soon. So the only real reason I could have for adding it is to make unit tests not reliant on the OS, but even that seems like overkill.

Changes I will make:
1. Remove the `BluetoothWiFiProvisioner` interface and make the corresponding private impl public
2. Remove the `bluetoothService` interface and instead create a private type `bluetoothServiceLinux` (which lets us get rid of the generic)

subsystems/provisioning/bluetooth/bluetooth_wifi_provisioner.go
```diff
@@ -0,0 +1,149 @@
...
+
+import (
+	"context"
+	"encoding/json"
+	"runtime"
+	"sync"
+
+	"errors"
+
+	errw "github.com/pkg/errors"
```
maxhorowitz, line 12: Copied this (including the `errw` prefix) from elsewhere in the Agent. The "github.com/pkg/errors" package has `errors.WithMessage(err, "some wrapping message")` and `errors.Errorf("message with arg %s", arg)` (which do not exist in the standard "errors" package).

subsystems/provisioning/bluetooth/bluetooth_wifi_provisioner.go
```diff
@@ -0,0 +1,149 @@
...
+		Strength    float64 `json:"strength"` // Inclusive range [0.0, 1.0], represents the % strength of a WiFi network.
+		RequiresPsk bool    `json:"requires_psk"`
+	} `json:"networks"`
+}
+
+func (awns *AvailableWiFiNetworks) ToBytes() ([]byte, error) {
+	return json.Marshal(awns)
+}
+
+// bluetoothWiFiProvisioner provides an interface for managing BLE (bluetooth-low-energy) peripheral advertisement on Linux.
```
maxhorowitz, line 39:
```suggestion
// BluetoothWiFiProvisioner provides an interface for managing BLE (bluetooth-low-energy) peripheral advertisement on Linux.
```

subsystems/provisioning/bluetooth/bluetooth_wifi_provisioner.go
maxhorowitz: Good point - will make more self documenting.

subsystems/provisioning/bluetooth/bluetooth_wifi_provisioner.go
maxhorowitz: Hmm - yes. That's right because if it enters that line it means we're requiring that value to be read properly.

subsystems/provisioning/bluetooth/bluetooth_wifi_provisioner.go
maxhorowitz: Yes, removed from there and passing it a layer below. Goal was to validate context was alive before passing something through a channel, which **should** be closed if context has been cancelled.

subsystems/provisioning/bluetooth/bluetooth_wifi_provisioner_utils.go
```diff
@@ -0,0 +1,44 @@
...
+// emptyBluetoothCharacteristicError represents the error which is raised when we attempt to read from an empty BLE characteristic.
+type emptyBluetoothCharacteristicError struct {
+	missingValue string
+}
+
+func (e *emptyBluetoothCharacteristicError) Error() string {
+	return fmt.Sprintf("no value has been written to BLE characteristic for %s", e.missingValue)
+}
+
+// waitForBLE is used to check for the existence of a new value in a BLE characteristic.
```
maxhorowitz, line 20:
```suggestion
// retryCallbackOnEmptyCharacteristicError retries the provided callback to at one second intervals as long as an expected error is thrown.
```

subsystems/provisioning/bluetooth/bluetooth_wifi_provisioner.go
```diff
@@ -0,0 +1,171 @@
...
+type Credentials struct {
+	Ssid           string
+	Psk            string
+	RobotPartKeyID string
+	RobotPartKey   string
+}
+
+// AvailableWiFiNetworks represent the networks that the device has detected (and which may be available for connection).
+type AvailableWiFiNetworks struct {
+	Networks []*struct {
```
ale7714, line 28: you can re-use this type https://github.com/viamrobotics/agent/blob/90b7f592f7fd5a1f29d1c3fde04dfc8d39e95ffe/subsystems/provisioning/definitions.go#L112

subsystems/provisioning/bluetooth/bluetooth_wifi_provisioner.go
```diff
@@ -0,0 +1,171 @@
...
+	"errors"
+
+	errw "github.com/pkg/errors"
+	"go.uber.org/multierr"
+	"go.viam.com/rdk/logging"
+	"go.viam.com/utils"
+)
+
+// Credentials represent the minimum required information needed to provision a Viam Agent.
+type Credentials struct {
```
ale7714, line 19: lets reuse https://github.com/viamrobotics/agent/blob/90b7f592f7fd5a1f29d1c3fde04dfc8d39e95ffe/subsystems/provisioning/definitions.go#L257 

These types will be the same in the BLE and the Hotspot flow

subsystems/provisioning/bluetooth/bluetooth_wifi_provisioner.go
```diff
@@ -0,0 +1,171 @@
...
+
+// BluetoothWiFiProvisioner provides an interface for managing BLE (bluetooth-low-energy) peripheral advertisement on Linux.
+type BluetoothWiFiProvisioner struct{}
+
+// Start begins advertising a bluetooth service that acccepts WiFi and Viam cloud config credentials.
+func (bwp *BluetoothWiFiProvisioner) Start(ctx context.Context) error {
+	if err := bwp.startAdvertisingBLE(ctx); err != nil {
+		return err
+	}
+	bwp.enableAutoAcceptPairRequest() // Async goroutine (hence no error check) which auto-accepts pair requests on this device.
```
ale7714, line 47: does this mean we won't provide helpful logging to the users? or that part will be implemented in `enableAutoAcceptPairRequest`?

subsystems/provisioning/bluetooth/bluetooth_wifi_provisioner.go
```diff
@@ -0,0 +1,171 @@
...
+) (*Credentials, error) {
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	if !requiresWiFiCredentials && !requiresCloudCredentials {
+		return nil, errors.New("should be waiting for either cloud credentials or WiFi credentials")
+	}
+	var ssid, psk, robotPartKeyID, robotPartKey string
+	var ssidErr, pskErr, robotPartKeyIDErr, robotPartKeyErr error
+	wg := &sync.WaitGroup{}
```
maxhorowitz, line 73:
```suggestion
	wg := sync.WaitGroup{}
```

subsystems/provisioning/bluetooth/bluetooth_wifi_provisioner.go
maxhorowitz: Logging is included in the implementation of enableAutoAcceptPairRequest. The stacked PR containing its impl is here: https://github.com/viamrobotics/agent/pull/69.

subsystems/provisioning/bluetooth/bluetooth_wifi_provisioner.go
maxhorowitz: Same as the above with regards to import cycles. Can use, but would need to pull that out. Otherwise, maybe we should just be moving this code into the existing `provisioning` package as its own file where it can also access those values?

subsystems/provisioning/bluetooth/bluetooth_wifi_provisioner.go
cheukt: I assume that using errw is just so we can migrate old code without too much hassle, but for new code you can use `fmt.Errorf`
```
// An easy way to create wrapped errors is to call [fmt.Errorf] and apply
// the %w verb to the error argument:
//
//	wrapsErr := fmt.Errorf("... %w ...", ..., err, ...)
```

maxhorowitz: Awesome. Didn't know it was doable that way. Will change to this, thanks.

subsystems/provisioning/bluetooth/bluetooth_wifi_provisioner.go
```diff
@@ -0,0 +1,171 @@
...
+					cancel()
+				}
+			},
+			wg.Done,
+		)
+	}
+	wg.Wait()
+	return &Credentials{
+		Ssid: ssid, Psk: psk, RobotPartKeyID: robotPartKeyID, RobotPartKey: robotPartKey,
+	}, multierr.Combine(ssidErr, pskErr, robotPartKeyIDErr, robotPartKeyErr)
```
cheukt, line 124: can use `errors.Join`

subsystems/provisioning/bluetooth/bluetooth_wifi_provisioner.go
```diff
@@ -0,0 +1,171 @@
...
+			func() {
+				if robotPartKey, robotPartKeyErr = retryCallbackOnExpectedError(
+					ctx, bwp.readRobotPartKey, &emptyBluetoothCharacteristicError{}, "failed to read robot part key",
+				); robotPartKeyErr != nil {
+					cancel()
+				}
+			},
+			wg.Done,
+		)
+	}
```
cheukt, line 119: this pattern feels odd to me - why are we starting 4 goroutines that's each calling 1 function in a loop? Something useful for context would be to know on a high level how each of the functions will work

subsystems/provisioning/bluetooth/bluetooth_wifi_provisioner.go
maxhorowitz: Moved into `provisioning` and reused this type.

subsystems/provisioning/bluetooth/bluetooth_wifi_provisioner.go
maxhorowitz: Moved into `provisioning` and reused this type.

subsystems/provisioning/bluetooth_wifi_provisioner.go
```diff
@@ -0,0 +1,192 @@
...
+	"sync"
+	"time"
+
+	"errors"
+
+	"go.viam.com/rdk/logging"
+	"go.viam.com/utils"
+)
+
+// bluetoothWiFiProvisioner provides an interface for managing BLE (bluetooth-low-energy) peripheral advertisement on Linux.
```
maxhorowitz, line 16:
```suggestion
// bluetoothWiFiProvisioner provides methods for managing BLE (bluetooth-low-energy) peripheral advertisement on Linux.
```

subsystems/provisioning/bluetooth_wifi_provisioner.go
```diff
@@ -0,0 +1,192 @@
...
+		}
+		select {
+		case <-ctx.Done():
+			return "", ctx.Err()
+		default:
+			time.Sleep(time.Second)
+		}
+		v, err := fn()
+		if err != nil {
+			if errors.As(err, &expectedErr) {
```
maxhorowitz, line 185: Note to self to fix this and manually test the error matching works.

subsystems/provisioning/bluetooth/bluetooth_wifi_provisioner.go
cheukt: I'm looking at https://github.com/viamrobotics/agent/pull/68/files so let me know if I'm missing something.

It seems possible to only have 1 goroutine that calls readSsid, readPsk, readRobotPartKeyID, and readRobotPartKey sequentially and then retry every second, is there a reason why not?

maxhorowitz:
@cheukt can't seem to get out of "review" mode on my PR which is preventing me from commenting on your comments.

> I'm looking at https://github.com/viamrobotics/agent/pull/68/files so let me know if I'm missing something. It seems possible to only have 1 goroutine that calls readSsid, readPsk, readRobotPartKeyID, and readRobotPartKey sequentially and then retry every second, is there a reason why not?
No reason not to do that, other than we will have to add additional state in this function for "skipping" a read if we've already received a value for it.

maxhorowitz (commented):
Comments for self.

subsystems/networking/bluetooth.go
```diff
@@ -0,0 +1,172 @@
...
+}
+
+func (bwp *bluetoothServiceLinux) readRobotPartKey() (string, error) {
+	return "", errors.New("TODO APP-7653: Implement helper methods to read SSID, passkey, robot part key ID, and robot part key" +
+		" values from BLE peripheral characteristics")
+}
+
+// NewBluetoothWiFiProvisioningServiceLinux returns a service which accepts credentials over bluetooth to provision a robot and its WiFi connection.
+func NewBluetoothWiFiProvisioningService(ctx context.Context, logger logging.Logger, name string) (*bluetoothServiceLinux, error) {
+	switch os := runtime.GOOS; os {
```
Otterverse, line 143: Generally we don't want to do per-OS logic at runtime. Go has a pretty clever mechanism (build constraints) for making different versions of a file (and therefore the functions it contains) build only under certain conditions. E.g. the "windows" build would include only the windows code, and the regular/linux file would include the linux versions. That keeps code cleaner and resulting binaries smaller as well.

See https://pkg.go.dev/cmd/go#hdr-Build_constraints for details, but feel free to ask me for more examples as we've used it quite a bit in RDK itself.

subsystems/networking/bluetooth.go
```diff
@@ -0,0 +1,166 @@
...
+					ssid != "" && psk != "" && robotPartKeyID != "" && robotPartKey != "" {
+					return
+				} else if requiresWiFiCredentials && ssid != "" && psk != "" {
+					return
+				} else if requiresCloudCredentials && robotPartKeyID != "" && robotPartKey != "" {
+					return
+				}
+
+				// Not ready to return (do not have the minimum required set of credentials), so sleep and try again.
+				time.Sleep(time.Second)
```
Otterverse, line 103: Don't use time.Sleep in agent/subsystems, it can cause health checks to fail. If you're inside a loop, use the mainLoopHealth/bgLoopHealth.Sleep(ctx) function instead

subsystems/networking/bluetooth.go
```diff
@@ -0,0 +1,166 @@
...
+	if !requiresWiFiCredentials && !requiresCloudCredentials {
+		return nil, errors.New("should be waiting for cloud credentials or WiFi credentials, or both")
+	}
+	var ssid, psk, robotPartKeyID, robotPartKey string
+	var ctxErr, ssidErr, pskErr, robotPartKeyIDErr, robotPartKeyErr error
+
+	wg := sync.WaitGroup{}
+	wg.Add(1)
+
+	utils.ManagedGo(func() {
```
Otterverse, line 51: I don't understand why you're backgrounding this, only to wg.Wait() right afterwards.

subsystems/networking/bluetooth.go
```diff
@@ -0,0 +1,166 @@
...
+				// Not ready to return (do not have the minimum required set of credentials), so sleep and try again.
+				time.Sleep(time.Second)
+			}
+		}
+	}, wg.Done)
+
+	wg.Wait()
+
+	return &userInput{
+		SSID: ssid, PSK: psk, PartID: robotPartKeyID, Secret: robotPartKey,
```
Otterverse, line 111: That's a lot of individual variables to have be shared into the lambda. Why not just declare &userInput{} itself and set it's contents inside the lambda?

Otterverse:
Closed in favor of #88

PR #65, opened 2025-02-18
[APP-7654]: Implement initializer function for creating a BLE peripheral with the required set of characteristics for BLE to WiFi provisioning.

maxhorowitz:


Otterverse:
Closed in favor of #88

PR #73, opened 2025-02-21
[APP-7692] Fix empty config errors on new install

Otterverse:
Default config wasn't being properly loaded on brand new installs with no cache. This only affects devices which are entirely offline at the time of install and rely on provisioning to get credentials (e.g. manufacturer preinstallation.)

Test build is here: http://packages.viam.com/temp/viam-agent-v0.14.1-rc1-aarch64

Manually copy it to an offline pi (with no existing /opt/viam) and run it with --install. Previously it would fail to enter provisioning. Now it should (as expected.)

### DIFF
--- a/subsystems/networking/scanning.go
+++ b/subsystems/networking/scanning.go
@@ -28,21 +28,31 @@ func (n *Networking) networkScan(ctx context.Context) error {
 		return errw.Errorf("cannot find hotspot interface: %s", n.Config().HotspotInterface)
 	}
 
+	state, reason, err := wifiDev.GetPropertyStateReason()
+	if err != nil {
+		return errw.Wrap(err, "getting wifi state and reason")
+	}
+
+	if state != gnm.NmDeviceStateDisconnected && state != gnm.NmDeviceStateActivated {
+		n.logger.Debugf("wifi device state: %s, reason: %s, skipping scan", state, reason)
+		return nil
+	}
+
 	prevScan, err := wifiDev.GetPropertyLastScan()
 	if err != nil {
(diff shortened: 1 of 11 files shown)


utils/config.go
```diff
@@ -195,7 +195,9 @@ func LoadConfigFromCache() (AgentConfig, error) {
 	//nolint:gosec
 	cacheBytes, err := os.ReadFile(cachePath)
 	if err != nil {
-		if !errors.Is(err, fs.ErrNotExist) {
+		if errors.Is(err, fs.ErrNotExist) {
+			return StackConfigs(&pb.DeviceAgentConfigResponse{})
+		} else {
```
Otterverse, line 200: This is the actual fix.

ale7714 (approved):

maxhorowitz (approved):
Thx!

subsystems/networking/scanning.go
```diff
@@ -28,21 +28,31 @@ func (n *Networking) networkScan(ctx context.Context) error {
...
+	state, reason, err := wifiDev.GetPropertyStateReason()
+	if err != nil {
+		return errw.Wrap(err, "getting wifi state and reason")
+	}
+
+	if state != gnm.NmDeviceStateDisconnected && state != gnm.NmDeviceStateActivated {
+		n.logger.Debugf("wifi device state: %s, reason: %s, skipping scan", state, reason)
+		return nil
+	}
+
```
Otterverse, line 40: This should help reduce the flood of "Scanning not allowed while unavailable" messages Eliot and Bill reported.

PR #77, opened 2025-03-03
[APP-7566]: Implement Bluetooth Agent Provisioning

maxhorowitz:
# Summary
This PR encompasses everything needed for adding bluetooth provisioning as a default provisioning method in the `viam-agent` (and is designed to run in parallel to the existing WiFi hotspot provisioning method).

# Setup
I want you to test as if you are a customer who has just received a machine in the mail. The machine will **just** have a `viam-agent` binary installation. It will not have WiFi credentials (and thus no internet connectivity).
## Machine setup
Using a Linux machine/laptop (_**that you are comfortable getting your `/opt/viam/*`, `/etc/viam/*` and other files tampered with**_) , pull down this branch. Because you will need to remove connectivity to emulate the user perspective, please do not SSH into a machine over WiFi (and don't use Ethernet because it can interfere with provisioning). Instead, **_test directly on the device_** (keyboard + mouse) needed.
### Uninstall Viam
First, you will need to remove any Viam-related stuff from your computer (to emulate the "newly-out-of-the-box" scenario our customers will be in). Please run the following from the repository root:
`sudo ./uninstall.sh`
### Remove local networks
You will then need to remove local networks from your machine to emulate the "offline-ness" that our customers will be dealing with. I've been running `nmcli con show` and subsequently `sudo nmcli con delete <name>` (for every network that could "interfere" with the provisioning flow).
### Preinstall the Viam Agent
Then, run the following (again from the repository root):
1. `make`
2. `sudo ./bin/viam-agent-custom-aarch64 --install`
3. `sudo /opt/viam/bin/viam-agent --debug`

I've been running the commands altogether as such:
`make && sudo ./bin/viam-agent-custom-aarch64 --install && sudo /opt/viam/bin/viam-agent --debug`
## Phone setup
Download the LightBlue app (available on both Android and iPhone). This is what we will use to communicate with the bluetooth service that is being advertised from the Linux machine.
# Testing Procedure
At this point, we are ready to test.
### Pairing
In the app, use the search bar to find the name of the Linux machine that you are currently testing against. It should be there. Pair with your machine (_one of the tests is to see if the pairing request is automatically accepted on the machine side, so I am hoping this part works!_). 

If it does not pair, check your screen for a six-digit pairing code and accept the request manually. This is a known limitation that I am working on fixing. The TLDR is it's complicated because of reliance on systemd bus signals which may be picked up and discarded as negligible elsewhere in the `viam-agent`.
### Bluetooth service used to "transmit" credentials
Once paired, look at the bluetooth service and nested characteristics available to us. You should be able to take the provided UUIDs and map them to the logs on your `viam-agent` machine:

There is an encoding that will be helpful for you to know here. Characteristics whose last 4 characters of their first 8 character sequence (preceding the `-`) will always be the following:
- `xxxx1111-...` is the encompassing service
- `xxxx2222-...` is the write-only characteristic for SSID 
- `xxxx3333-...` is the write-only characteristic for passkey
- `xxxx4444-...` is the write-only characteristic for part ID 
- `xxxx5555-...` is the write-only characteristic for part secret 
- `xxxx6666-...` is the write-only characteristic for app address
- `xxxx7777-...` is the read-only characteristic for nearby available WiFi networks that the machine has detected 

```
2025-02-25T20:41:24.245Z	DEBUG	viam-agent	networking/bluetooth.go:507	BlueZ version (5.66) meets the requirement (5.66 or later)
2025-02-25T20:41:24.246Z	DEBUG	viam-agent	networking/bluetooth.go:43	Bluetooth peripheral service UUID: 79ff1111-4f38-44b9-b3b5-78fb7e14757e
2025-02-25T20:41:24.246Z	DEBUG	viam-agent	networking/bluetooth.go:45	WiFi SSID can be written to the following bluetooth characteristic: db8f2222-baea-452f-bad0-62b440f98161
2025-02-25T20:41:24.246Z	DEBUG	viam-agent	networking/bluetooth.go:47	WiFi passkey can be written to the following bluetooth characteristic: 06af3333-4442-42e5-9e6a-1729ffa1378d
2025-02-25T20:41:24.246Z	DEBUG	viam-agent	networking/bluetooth.go:49	Robot part key ID can be written to the following bluetooth characteristic: 26d24444-161f-4474-a4a5-e83c563ac52b
2025-02-25T20:41:24.246Z	DEBUG	viam-agent	networking/bluetooth.go:51	Robot part key can be written to the following bluetooth characteristic: 12db5555-ca13-4a0b-879a-c76fb1542fe1
2025-02-25T20:41:24.246Z	DEBUG	viam-agent	networking/bluetooth.go:53	Viam app address can be written to the following bluetooth characteristic: 9eaa6666-14f0-47a9-b69f-a093f8358b18
2025-02-25T20:41:24.246Z	DEBUG	viam-agent	networking/bluetooth.go:55	Available WiFi networks can be read from the following bluetooth characteristic: a8ee7777-2496-485a-b0ea-dc63a1122f1
```
The LightBlue interface is pretty easy to follow. You may need to change LightBlue messages from hex or binary to utf-8 string. Once confirmed you're in utf-8 string mode, you can write individual messages to the SSID, passkey, part ID, part secret, and app address "characteristics" by clicking "write value." Similarly, you can read from the available WiFi networks "characteristic" by clicking "read value." Any value written will get "sucked in" to the `viam-agent` provisioning flow. **_You can check this in the `viam-agent` logs_**. Once all values are submitted, the provisioning loop should close out the BT connection, connect to WiFi, retrieve its cloud config, and should start up a `viam-server` (thus ending the provisioning loop).
## Cases
1. Hotspot provisioning flow (i.e. check the captive portal flow still works)
2. Bluetooth provisioning flow (i.e. check the LightBlue flow works)


maxhorowitz:
@Otterverse I'm the worst - I needed to create a new PR because when I force pushed from my other Pi (which was actually on `origin/main` the git commit history wiped itself). The other PR still exists (it's closed), and I am copying your review here:

> I apologize in advance for how long this review is. I'm leaving a summary here at the top of (most) of the important things, so it's easier to discuss (and check off as you go) hopefully. But it will overlap a lot with the inline comments too.

As always, hit me up in slack to discuss things whenever. Plenty of this is non-obvious, as you're deep in the guts of the most complicated part of Agent here.

Background go routines/threads need to be used only when/where required, and must be fully managable. Remember that calls to start/stop can come in at any point in the flow, so everything must be able to exit cleanly and quickly. No need to background threads if you're just going to wait for them in the same fuction. Just handle things linearly to keep it simpler.

Need to be careful about data races, and protect any data that can be changed from another thread with locks. Note that even if it's not explicitly in a thread, outside calls from start/stop/update can happen at any point, so everything in those paths has to be race-safe.

BT characteristics (and their associated UUIDs) are a bunch of indivudual variables resulting in ~1/3 of bluetooth.go being very repetitive boilerplate. Should get them arranged into a simpler data structure, with a map to use directly, or a true structure with (generalized) getter/setter methods. E.g. calls should be able to look like val, err := bt.GetCharacteristic("ssid") or bt.SetCharacteristic("ssid", myVal) and be human readable. Look at networkState and connState for hints.

As discussed in slack, the 20 character limit for charcteristic could be a real problem. SSIDs and PSKs can be longer than that by themselves. And scan results may contains dozens of networks.

Need a new config setting to disable bluetooth.

If agent starts on a device without bluetooth (or otherwise unsupport version/etc.), it needs to detect this and quietly do nothing (beyond an initial log.) Likely best to integrate with above, and just have a "temporary" flag that disables it until a full restart.

BLE needs to integrate with the health checking. Any backgrounded routines need to report health status regularly using health.healthySleep() (add a new one on Networking{} to track ble health.)

Ideally, this should be abstracted just like the portal is. E.g. whenever start/stopPortal() is called, start/stopBLE() is called, and otherwise works the same. They should even use the same inputChan, so the outer provisioning code only has to listen to that one channel.

Network scans (and other info) are updated in real time. This is why they're managed in their own structures like n.connState(). BLE needs to use those and not just pass in a static list one time at startup.

subsystems/networking/definitions.go
```diff
@@ -88,12 +88,12 @@ func (n *network) getInfo() NetworkInfo {
...
-	Security  string
-	Signal    int32
-	Connected bool
-	LastError string
+	Type      string `json:"type,omitempty"`
+	SSID      string `json:"ssid"`
+	Security  string `json:"sec"`
+	Signal    int32  `json:"sig"`
+	Connected bool   `json:"conn,omitempty"`
+	LastError string `json:"lastErr,omitempty"`
```
maxhorowitz, line 96: Max note to self to not use shorthand.

subsystems/networking/definitions.go
```diff
@@ -88,12 +88,12 @@ func (n *network) getInfo() NetworkInfo {
 }
 
 type NetworkInfo struct {
```
maxhorowitz, line 90: @Otterverse you mentioned to use a different, smaller struct for this use case. In a previous PR, Ale suggested reusing the `NetworkInfo` type specifically. You own this codebase, so lmk if you want something split out. My opinion is to use `NetworkInfo` because its in the same package and we're using it the prescribed way.

Otterverse: IF you can use it without changing it, that'd be fine. But you added shorthand tags, which does change the existing behavior (in theory... it may not actually harm anything currently.) You're also intentionally trying to skip several fields, so a more compact struct specifically for your parsing would probably be better regardless.

That said, unless you've solved the 20 byte limit, this doesn't matter, as json isn't going to be compact enough. If you HAVE solved it, then the question is "what's the new limit?" If there's still a tight limit (I read somewhere that BLE characteristics are 512 bytes long in the standards, not 20) then bytepacking might make more sense than json. If there's a more lax limit via some new method though, then you could possibly use the stucture as-is, without the new tags.

subsystems/networking/networking.go
```diff
@@ -243,7 +247,7 @@ func (n *Networking) Stop(ctx context.Context) error {
 
 	n.logger.Infof("%s subsystem exiting", SubsysName)
 	if n.connState.getProvisioning() {
-		err := n.stopProvisioning()
+		err := n.StopProvisioning()
```
maxhorowitz, line 250:
```suggestion
		err := n.stopProvisioning()
```

subsystems/networking/connstate.go
```diff
@@ -99,15 +108,29 @@ func (c *connectionState) getConfigured() bool {
...
+	switch c.provisioningMode {
+	case none:
+		return false
+	case hotspotOnly, bluetoothOnly, hotspotAndBluetooth:
+		return true
+	}
+	return false
+}
+
+// setProvisioningMode sets the provisioning mode
```
maxhorowitz, line 124:
```suggestion
// setProvisioningMode sets the provisioning mode.
```

subsystems/networking/connstate.go
```diff
@@ -99,15 +108,29 @@ func (c *connectionState) getConfigured() bool {
 	return c.configured
 }
 
-func (c *connectionState) setProvisioning(mode bool) {
+// getProvisioning returns true if in an active provisioning mode.
```
maxhorowitz, line 111:
```suggestion
// getProvisioning returns true if in provisioning mode, and the time of the last state change.
```

subsystems/networking/bluetooth.go
```diff
@@ -0,0 +1,740 @@
...
+		// Used to refresh the available WiFi networks.
+		requestAvailableWiFiNetworks: requestAvailableWiFiNetworksFn,
+	}
+
+	return bsl, &health, nil
+}
+
+// ---------------------------------------------------------------------------------------
+// ----------------------------------- EXTERNAL METHODS ------------------------------------
+// ---------------------------------------------------------------------------------------
```
maxhorowitz, line 121:
```suggestion
```

subsystems/networking/bluetooth.go
```diff
@@ -0,0 +1,737 @@
...
+	}
+	bsl.advActive = false
+	bsl.logger.Debug("Stopped advertising bluetooth service.")
+
+	bsl.cancelFunc()
+	bsl.workers.Wait()
+	return nil
+}
+
+// ---------------------------------------------------------------------------------------
```
maxhorowitz, line 159: Note: I can remove these extra visual dividers - added so it was easier for us to digest in a sync code review.

subsystems/networking/bluetooth.go
```diff
@@ -0,0 +1,737 @@
...
+	parts := strings.Split(path, "/")
+	if len(parts) < 4 {
+		return ""
+	}
+
+	// Extract last part and convert underscores to colons
+	macPart := parts[len(parts)-1]
+	mac := strings.ReplaceAll(macPart, "_", ":")
+	return mac
+}
```
maxhorowitz, line 583: Max note to self did not fix this yet.

subsystems/networking/networking.go
```diff
@@ -54,6 +54,10 @@ type Networking struct {
 	grpcServer *grpc.Server
 	portalData *portalData
 
+	// bluetooth
+	bluetoothService bluetoothService
+	bluetoothHealth  *health
```
maxhorowitz, line 59: Nott to self to ask james about health

subsystems/networking/connstate.go
```diff
@@ -99,15 +108,29 @@ func (c *connectionState) getConfigured() bool {
...
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	switch c.provisioningMode {
+	case none:
+		return false
+	case hotspotOnly, bluetoothOnly, hotspotAndBluetooth:
+		return true
+	}
+	return false
+}
```
maxhorowitz, line 122: This should be all be either on or off

subsystems/networking/networkmanager.go
```diff
@@ -173,6 +174,32 @@ func (n *Networking) StartProvisioning(ctx context.Context, inputChan chan<- use
...
+	} else if hotspotErr == nil && bluetoothErr != nil {
+		n.connState.setProvisioningMode(hotspotOnly)
+		return bluetoothErr
+	} else if hotspotErr != nil && bluetoothErr == nil {
+		n.connState.setProvisioningMode(bluetoothOnly)
+		return hotspotErr
+	} else {
+		n.connState.setProvisioningMode(hotspotAndBluetooth)
+		return nil
+	}
```
maxhorowitz, line 198: Switch statement goes away. It will be just set to true once we get rid of enumeration

subsystems/networking/networkmanager.go
```diff
@@ -203,13 +256,34 @@ func (n *Networking) StopProvisioning() error {
...
 	err := n.stopPortal()
 	err2 := n.deactivateConnection(n.Config().HotspotInterface, n.Config().HotspotSSID)
 	if errors.Is(err2, ErrNoActiveConnectionFound) {
 		return err
 	}
-	return errors.Join(err, err2)
+	if err := errors.Join(err, err2); err != nil {
+		return err
+	}
+	n.logger.Info("Stopped hotspot provisioning mode.")
```
maxhorowitz, line 275: Make debug loglines for this and for bluetooth.

subsystems/networking/networkmanager.go
```diff
@@ -190,8 +217,34 @@ func (n *Networking) StartProvisioning(ctx context.Context, inputChan chan<- use
...
+	if n.bluetoothService != nil {
+		return nil
+	}
+	deviceName := fmt.Sprintf("%s.%s.%s", n.Config().Manufacturer, n.Config().Model, n.Config().FragmentID)
+	bt, health, err := newBluetoothService(n.logger, deviceName, n.getVisibleNetworks)
+	if err != nil {
+		return err
+	}
+	n.bluetoothService = bt
+	n.bluetoothHealth = health
```
maxhorowitz, line 247: move to `init`

subsystems/networking/bluetooth.go
```diff
@@ -0,0 +1,737 @@
...
+		return errors.New("should be waiting for cloud credentials or WiFi credentials, or both")
+	}
+
+	var ssid, psk, robotPartKeyID, robotPartKey, appAddress string
+	var e *emptyBluetoothCharacteristicError
+	for {
+		if ctx.Err() != nil {
+			return ctx.Err()
+		}
+		select {
```
maxhorowitz, line 337: I can remove this its not needed.

subsystems/networking/bluetooth.go
```diff
@@ -0,0 +1,737 @@
...
+	bsl.advActive = true
+	bsl.adv = defaultAdvertisement
+	bsl.logger.Debugf("Started advertising bluetooth service with UUID: %s.", serviceUUID.String())
+
+	return nil
+}
+
+// run spawns three goroutines, one to refresh available WiFi networks, another to listen for user input, and the last to listen for bluetooth
+// pairing requests.
+func (bsl *bluetoothServiceLinux) run(ctx context.Context, requiresCloudCredentials, requiresWiFiCredentials bool, inputChan chan<- userInput) {
```
maxhorowitz, line 206: Mark this as unhealthy if we are returning early - something needs to know that this is no longer "actively bluetooth-ing"

subsystems/networking/bluetooth.go
```diff
@@ -0,0 +1,737 @@
...
+				ssid != "" && psk != "" && robotPartKeyID != "" && robotPartKey != "" && appAddress != "" {
+				break
+			} else if requiresWiFiCredentials && ssid != "" && psk != "" {
+				break
+			} else if requiresCloudCredentials && robotPartKeyID != "" && robotPartKey != "" && appAddress != "" {
+				break
+			}
+
+			// If we haven't received all required credentials, try again a second later.
+			if !bsl.health.Sleep(ctx, time.Second) {
```
maxhorowitz, line 394: The purpose of the healthy sleep is to be able to exit a sleep before the entire duration because context was cancelled during the sleep.

** bail out of the sleep **

subsystems/networking/bluetooth.go
```diff
@@ -0,0 +1,737 @@
...
+		return err
+	}
+	if err := bsl.listenForPropertyChanges(conn); err != nil {
+		return err
+	}
+
+	// Listen for properties changed events (includes bluetooth pairing requests).
+	signalChan := make(chan *dbus.Signal, 25)
+	conn.Signal(signalChan)
+	for {
```
maxhorowitz, line 601: refactor so its clearer

subsystems/networking/bluetooth.go
```diff
@@ -0,0 +1,737 @@
...
+				continue
+			}
+			// Check if the signal is from a BlueZ device
+			if len(signal.Body) < 3 {
+				continue
+			}
+
+			iface, ok := signal.Body[0].(string)
+			if !ok || iface != "org.bluez.Device1" {
+				continue
```
maxhorowitz, line 619: see if i can naturally filter this

Otterverse:
From our discussion about bytepacking, my suggestion for the network list is to have the first byte be metadata, and the rest be the SSID, terminated by a null byte. For the "metadata" that would be signal and security. Use the high bit of that byte to indicate "secure" network, and the remaining 7 bits for 0-128 signal strength.

So a Network called "foobar" with 50% signal strength and security enabled would be something like this (did the full return-trip to check my own understanding.)

```
ssid := "foobar"
signal := uint8(50) // can't be more than 127, use 0-100)
secure := true

meta := byte(signal)
if secure {
	meta = meta | (1 << 7)
}

list := []byte{meta}
list = append(list, []byte(ssid)...)
list = append(list, 0x0) // separator/terminator

fmt.Printf("HEX: % x\n", list)

newMeta := list[0]
newSecure := uint8(newMeta) > 127
newSignal := newMeta &^ byte(1<<7)
newSsid := string(list[1:])

fmt.Printf("SSID: %s, Signal: %d, Secure: %t\n", newSsid, newSignal, newSecure)
```
Adding more entries (and splitting them afterwards) is left as an exercise to the reader.

subsystems/networking/bluetooth.go
```diff
@@ -0,0 +1,714 @@
...
+	listenForCredentialsHealth        *health
+	listenForPairRequestHealth        *health
+
+	// Stopping/restarting an existing bluetooth service requires the following state variables.
+	trustedDevices              []string
+	bluezAgentRegistered        bool
+	listeningForPropertyChanges bool
+	adv                         *bluetooth.Advertisement
+	advActive                   bool
+	characteristicsByName       map[string]*bluetoothCharacteristicLinux[*string]
```
maxhorowitz, line 64: Please do not review this change.

subsystems/networking/bluetooth.go
```diff
@@ -0,0 +1,714 @@
...
+
+		// Used to manage whether provisioning completes successfully.
+		provisioningComplete: make(chan struct{}),
+	}
+
+	return bsl, nil
+}
+
+// Start begins advertising a bluetooth service that advertises nearby networks and acccepts WiFi and/or Viam cloud config credentials.
+func (bsl *bluetoothServiceLinux) start(
```
maxhorowitz, line 126: Need state external to the bluetoothServiceLinux type which will **refrain** from calling start or stop if the bluetoothServiceLinux is nil to begin with.

subsystems/networking/bluetooth.go
```diff
@@ -0,0 +1,714 @@
...
+	characteristicsByName       map[string]*bluetoothCharacteristicLinux[*string]
+
+	// Below callback functions are used to retrieve visible networks and to write that list of available networks to
+	// a bluetooth characteristic. The bluetooth characteristic, "Available WiFi Networks", is read-only from the
+	// perspective of a client.
+	requestAvailableWiFiNetworks func() []NetworkInfo
+	refreshAvailableWiFiNetworks func(p []byte) (e error)
+
+	// Channel remains open until we have successfully completed provisioning.
+	provisioningComplete chan struct{}
```
maxhorowitz, line 73: Remove this.

subsystems/networking/bluetooth.go
```diff
@@ -0,0 +1,698 @@
...
+			break
+		} else if requiresCloudCredentials && !requiresWiFiCredentials && robotPartKeyID != "" && robotPartKey != "" && appAddress != "" {
+			break
+		}
+
+		// If we haven't received all required credentials, try again a second later.
+		if !bsl.listenForCredentialsHealth.Sleep(ctx, time.Second) {
+			return nil // Return nil instead of a canceled context error to gracefully exit.
+		}
+		continue
```
Otterverse, line 402:
```suggestion
```
nit, no need to continue at the end of a loop, that always happens

subsystems/networking/bluetooth.go
```diff
@@ -0,0 +1,698 @@
...
+	err = conn.BusObject().Call("org.freedesktop.DBus.AddMatch", 0, matchRule).Err
+	if err != nil {
+		return fmt.Errorf("failed to add DBus match rule: %w", err)
+	}
+	bsl.listeningForPropertyChanges = true
+	bsl.logger.Debug("Listening for property changes (bluetooth pairing requests) on system D-bus.")
+
+	// Listen for properties changed events (includes bluetooth pairing requests).
+	signalChan := make(chan *dbus.Signal, 25)
+	conn.Signal(signalChan)
```
Otterverse, line 452: Low priority: You're wrapping this entire function itself in a background routine. In theory, this function should do most of this inline (so these errors can get returned quickly during this "setup" phase) and then only background the loop itself below that will actually be waiting for stuff to happen.

subsystems/networking/networkmanager.go
```diff
@@ -190,8 +206,17 @@ func (n *Networking) StartProvisioning(ctx context.Context, inputChan chan<- use
...
 	}
+	n.logger.Info("Hotspot provisioning set up successfully.")
+	return nil
+}
 
-	n.connState.setProvisioning(true)
+// startProvisioningBluetooth should only be called by 'StartProvisioning' (to ensure opMutex is acquired).
+func (n *Networking) startProvisioningBluetooth(ctx context.Context, inputChan chan<- userInput,
+) error {
+	if err := n.bluetoothService.start(ctx, true, true, inputChan); err != nil {
```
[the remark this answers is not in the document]
Otterverse: Also, you have two arguments here that you're hardcoding to true. Looks like they say what credential are needed. Don't you have to change the inputs here to match reality? E.g. how does it know when the viam config is already done?

Note that there are bits of this on the state machine above. `n.connState.GetConnected()` an `n.connState.GetConfigured()` Should probably use those dynamically deeper in the code, rather than passing in values only at startup.

subsystems/networking/bluetooth.go
```diff
@@ -0,0 +1,698 @@
...
+	cUUID := bluetooth.NewUUID(uuid.New()).Replace16BitComponent(encoding)
+	bsl.logger.Debugf("%s can be read from the following bluetooth characteristic: %s", cName, cUUID.String())
+	c := &bluetooth.Characteristic{}
+
+	// refreshAvailableWiFiNetworks represents the callback function used
+	// to dynamically update an otherwise read-only bluetooth characteristic.
+	bsl.refreshAvailableWiFiNetworks = func(bs []byte) error {
+		_, err := c.Write(bs)
+		return err
+	}
```
Otterverse, line 606: This is creating this callback for ALL read-only characteristics?

Otterverse: Wait, I see the networklist is the only read-only one. I'd assumed you were passing the "needWiFi" and "needConfig" values to the mobile device this way,, the same way the wifi provisioning does (so the UI doesn't prompt for unneeded fields.)

Looks like you're also not sharing the fragment_id? I'm pretty sure that one is critical for provisioning to work.

subsystems/networking/bluetooth.go
maxhorowitz: Did this with `noBS` (which sounds pretty funny in retrospect 😂 ) - it means "no bluetooth service"

subsystems/networking/bluetooth.go
```diff
@@ -0,0 +1,698 @@
...
+}
+
+// initializeBluetoothService performs low-level system configuration to enable bluetooth advertisement.
+func initializeBluetoothService(deviceName string, characteristics []bluetooth.CharacteristicConfig,
+) (bluetooth.UUID, *bluetooth.Advertisement, error) {
+	adapter := bluetooth.DefaultAdapter
+	if err := adapter.Enable(); err != nil {
+		return [4]uint32{}, nil, fmt.Errorf("failed to enable bluetooth adapter: %w", err)
+	}
+	serviceUUID := bluetooth.NewUUID(uuid.New()).Replace16BitComponent(0x1111)
```
Otterverse, line 621: Question: I see these UUIDs getting used all over, but uuid.New() is generating NEW uuids every run of the software. Is that expected/normal? Or are you suppose to define the full thing once/permanantly for a "known service" or anything like that? Just feels weird to generate new UUIDs for consistent elements.

maxhorowitz:
@Otterverse wanted to call out that I tested this with a `soft-block` on my onboard bluetooth and the behavior was correct - it bypassed all of the bluetooth service stuff without erroring, and permanently set the `noBS` field to `true` so that those `startBluetoothProv` and `stopBluetoothProv` calls would return before doing anything.

subsystems/networking/bluetooth.go
```diff
@@ -0,0 +1,698 @@
...
+	for _, cmd := range versionCmds {
+		versionOutput.Reset() // Clear buffer
+		cmdParts := strings.Fields(cmd)
+		execCmd := exec.Command(cmdParts[0], cmdParts[1:]...) //nolint:gosec
+		execCmd.Stdout = &versionOutput
+		err = execCmd.Run()
+		if err == nil {
+			break // Found a valid command
+		}
+	}
```
Otterverse, line 671: Somewhat minor for now, but I do think it'd be better to test this via D-BUS (and during initBluetooth()) in order to really validate that you're talking to bluez in the expected way. Running commands may not always work as expected, and even if they're installed, it doesn't mean the service is actually running.

Also, there's an easier way to just run a one-shot command without setting up your own buffers or anything.
```
cmd := exec.Command("bluetoothctl", "--version")
output, err := cmd.CombinedOutput()
```

subsystems/networking/bluetooth.go
```diff
@@ -0,0 +1,698 @@
...
+	serviceUUID, defaultAdvertisement, err := initializeBluetoothService(
+		bsl.deviceName,
+		[]bluetooth.CharacteristicConfig{
+			ssidCharacteristicConfig,
+			pskCharacteristicConfig,
+			robotPartKeyIDCharacteristicConfig,
+			robotPartKeyCharacteristicConfig,
+			appAddressCharacteristicConfig,
+			availableWiFiNetworksCharacteristicConfig,
+		},
```
Otterverse, line 199: I think this is missing some key fields. Ideally BT provisioning should provide the same info (and therefore UI flow) as wifi. Look at line 40 of grpc.go for the list there that's sent to the mobile device. Manufacturer and Model are probably okay to skip, but FragmentID is critical for provisioning. Similarly "HasSmartMachineCredentials" and "IsOnline" values inform the mobile UI what fields to show (or not) when asking the user for input. There's also a list of errors provided, so when, say, a bad wifi password is provided (or something else goes wrong) it can be reported to the user.

subsystems/networking/setup.go
```diff
@@ -151,3 +152,17 @@ func (n *Networking) enableWifi(ctx context.Context) error {
 		}
 	}
 }
+
+func (n *Networking) initBluetooth() error {
+	if n.bluetoothService != nil {
+		return nil
+	}
+	deviceName := fmt.Sprintf("%s-%s-%s", n.Config().Manufacturer, n.Config().Model, n.Config().FragmentID)
```
Otterverse, line 160: Are there constraints on the device name? Manufacturer and Model could both be rather long, and fragment ID is a full 32-byte UUID, so that'll look really weird on a phone. Also, all robots with the same config would look the same, so you couldn't tell them apart. I would suggest using the same pattern as the wifi. That is, prefix+hostname: 

`n.Config().HotspotPrefix + "-" + strings.ToLower(hostname)`
You can get the hostname from `os.Hostname()`

Otterverse (changes requested):
Lot more stuff, sorry. Around half is kinda minor though. Note that I didn't add any comments around the characteristics/map/struct thing, as you said you'd work on that already.

We'll sync next week when I'm back, so you've got some time here. When we meet up, I'll probably ask you to walk me through (or provide instructions/links) for what to install on my phone and how to use it for testing this going forward.

Thanks!

subsystems/networking/networkmanager.go
```diff
@@ -190,8 +206,21 @@ func (n *Networking) StartProvisioning(ctx context.Context, inputChan chan<- use
...
 
-	n.connState.setProvisioning(true)
+// startProvisioningBluetooth should only be called by 'StartProvisioning' (to ensure opMutex is acquired).
+func (n *Networking) startProvisioningBluetooth(ctx context.Context, inputChan chan<- userInput,
+) error {
+	if n.noBS {
+		return nil
+	}
+	if err := n.bluetoothService.start(ctx, true, true, inputChan); err != nil {
+		n.noBS = true
```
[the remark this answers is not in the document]
maxhorowitz: That was my design, yes. Do you have an opinion?

subsystems/networking/bluetooth.go
```diff
@@ -0,0 +1,698 @@
...
+
+// newBluetoothService returns a service which accepts credentials over bluetooth to provision a robot and its WiFi connection.
+func newBluetoothService(
+	logger logging.Logger,
+	deviceName string,
+	requestAvailableWiFiNetworksFn func() []NetworkInfo,
+) (bluetoothService, error) {
+	if err := validateBlueZVersion(); err != nil {
+		return nil, fmt.Errorf("system requisites not met: %w", err)
+	}
```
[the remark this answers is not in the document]
maxhorowitz: Will do that. Agreed.

subsystems/networking/bluetooth.go
```diff
@@ -0,0 +1,698 @@
...
+	bsl.listenForCredentialsHealth = h
+	h = &health{}
+	h.MarkGood()
+	bsl.listenForPairRequestHealth = h
+
+	// Start goroutine to update the list of available WiFi networks.
+	bsl.workers.Add(1)
+	utils.ManagedGo(
+		func() {
+			if err := updateAvailableWiFiNetworks(ctx, bsl); err != nil {
```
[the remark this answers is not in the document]
maxhorowitz: I think they need to happen synchronously so that we can be sure to not get a nil pointer exception. Am I correct to say that?

subsystems/networking/bluetooth.go
```diff
@@ -0,0 +1,698 @@
...
+				bsl.logger.Errorw("failed to update available WiFi networks", "error", err)
+				bsl.updateAvailableWiFiNetworksHealth.MarkBad()
+
+				// Only cancel on failures. Failures indicate we've hit some exception and are
+				// unable to write the latest, available WiFi networks to our bluetooth service.
+				bsl.cancelFunc()
```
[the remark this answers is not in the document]
maxhorowitz: Those are not going away, those are used to signal to the other async goroutines that they should return. We do handle the context cancelled error inside of each as a "return nil" case ... i.e. swallow it and return gracefully.

I don't see any scenario where we could remove the shared cancel function held on the `bsl` struct - because thats the only way to signal shutdown of these three goroutines.

subsystems/networking/bluetooth.go
maxhorowitz: Thanks - sorry this was leftover and didn't remove 😄

subsystems/networking/networking.go
```diff
@@ -54,6 +54,10 @@ type Networking struct {
 	grpcServer *grpc.Server
 	portalData *portalData
 
+	// bluetooth
+	noBS             bool
```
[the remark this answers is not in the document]
maxhorowitz: Sorry about that 😅

subsystems/networking/bluetooth.go
maxhorowitz: Two important pieces here:

### 1 - the passing of fields "needs WiFi" and "needs config" values to the mobile app. 

I can do this to mirror what you have for the captive portal. Should be an easy addition - and I do agree that its necessary.

### 2 -  the fragment ID. 

I think there are two very different provisioning scenarios here,  so I want to make sure we're not talking past each other.

Scenario A: customer (i.e. Canyon Runner) sends devices that are pre-flashed with their company's fragment ID. Every customer gets the same fragment ID, and it doesn't get "set" by the end user. This is the most important / frequent use case according to Eliot. And in that case, the only things that get sent over should be WiFi credentials. I don't know this code path well at all - but the machine **should** be making a request directly to the cloud with a fragment ID only, and the cloud should create a new machine and return its cloud config to the robot. 

Question for you: _does no fragment ID make sense?_

Scenario B: someone wants to set up their machine that already exists in the cloud. In that case we want to send the credentials from the phone to the machine (included in this PR).

Deciding whether we are in Scenario A or Scenario B should happen outside of the `bluetooth.go` code - I think it should be a provisioning decision, and that's why I included the [currently unused] boolean args for "needs WiFi credentials" and "needs cloud credentials" - essentially deferring to the caller to dictate which scenario we are in.

Question for you: _does giving all control to the caller make sense?_

PR #78, opened 2025-03-04
[RSDK-7007] Add Manifest Generation in CI

Otterverse:
Add manifest generation to agent builds, to support incoming work in App.


.github/workflows/workflow.yaml
```diff
@@ -50,13 +50,18 @@ jobs:
...
     - uses: google-github-actions/setup-gcloud@v2
-    - name: Upload binaries to GCS
-      run: gsutil -h "Cache-Control:no-cache" cp bin/viam-agent-* gs://packages.viam.com/temp/prerelease/ #TODO use real target after live testing
-    - name: "Upload install scripts to GCS"
+    - name: Upload binaries to GCS (Release)
       if: github.event_name == 'release'
-      run: gsutil -h "Cache-Control:no-cache" cp preinstall.sh install.sh uninstall.sh gs://packages.viam.com/temp/prerelease/ #TODO use real target after live testing
+      run: |
+        gsutil -h "Cache-Control:no-cache" cp bin/viam-agent-* install.sh uninstall.sh preinstall.sh gs://packages.viam.com/apps/viam-agent/
+        gsutil cp etc/viam-agent-*.json gs://packages.viam.com/apps/viam-subsystems/
```
abe-winter, line 62: the second `gsutil cp` line is duplicated, could be moved to its own step so it can be shared?

only do this if it makes this clearer from your perspective though -- it's a style nit

manifest.sh
```diff
@@ -0,0 +1,45 @@
...
+
+# Determine if this is a prerelease
+if [[ "$VERSION" == *"-dev"* ]]; then
+    BINNAME="prerelease/$BINARY_NAME"
+else
+    BINNAME="$BINARY_NAME"
+fi
+
+# Set platform based on architecture in binary name
+if [[ "$BINARY_NAME" == *"x86_64"* ]]; then
```
abe-winter, line 23: I say this in the tone of the trailer for a horror movie but remember, windows is coming

linux/amd64 isn't the only amd64 anymore

I think you need to do one of:
- handle `*windows-x86_64` first, and make sure `*x86_64` is an else case
- always include OS in the filename

abe-winter (approved):
LGTM, guessing you will need to test e2e, also see my note about windows support in filenames

manifest.sh
abe-winter: you're assuming I haven't created an entire parallel manifest system on my branch that will lead to terrible merge conflicts

PR #79, opened 2025-03-12
RSDK-9990 - Use SOCKS proxy in agent

cheukt:
direct grpc dials will already use the socks proxy if the env var exists, so making changes to the other connections create by the agent.

the fallback dialer prioritizes using wifi, but will use the socks proxy as a fallback. Since these connections are one-off, I don't think we'll need to make any large changes to DownloadFile for RSDK-9991 (where we want to proactively switch to wifi/ethernet if available).

tested with a locally built agent and seeing that the viam-server was successfully downloaded (after 8 mins or so!). Tested with a variety of env vars and available connections, but let me know if more testing is needed!

still have to bump goutils in agent once goutils change is merged

see https://github.com/viamrobotics/goutils/pull/414 for main changes


Otterverse (approved):
LGTM (after moving to "real" goutils version with stuff included.)

PR #82, opened 2025-03-13
APP-7841 symlink agent binary in MSI

abe-winter:
## What changed
- When installing agent-for-windows from MSI, install viam-agent.exe in cache/, symlink in bin/.
    - Note there is no native way to do symlinks in wix, *and* the ServiceInstall directive needs to point at a File entry that wix knows about. That's why the wix xml file is kind of odd now.
## Why
agent.bat and the MSI were putting the binary directly in /opt/viam/bin, which makes it impossible for agent to automatically self-upgrade.

(Because on windows the file for a running executable is locked).
## Testing
- [x] CI yaml can build an MSI https://github.com/viamrobotics/agent/actions/runs/13850097372
- [ ] MSI produced by CI yaml creates a working agent installation
    - [ ] including on device not in developer mode (because mklink is a privileged command)
- [ ] what happens when this is installed over existing msi install?

### DIFF
--- a/agent.wxs
+++ b/agent.wxs
@@ -17,11 +17,7 @@
                             <fire:FirewallException Id="AllowAllUDP" Name="viam-agent" Profile="all" Protocol="tcp" Scope="any" />
                         </Component>
                     </Directory>
-                    <Directory Id="BINFOLDER" Name="bin">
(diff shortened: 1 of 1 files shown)


Otterverse (approved):
LGTM

ale7714 (approved):

PR #83, opened 2025-03-19
Fix go build on darwin

ale7714:
Main motivation is my local dev exp. Mainly i want to be able to run make lint 🥲

### DIFF
--- a/cmd/viam-agent/main_darwin.go
+++ b/cmd/viam-agent/main_darwin.go
@@ -1,0 +1,28 @@
+package main
+
+import (
+	"context"
+	"os"
+
+	"github.com/viamrobotics/agent"
+	"github.com/viamrobotics/agent/utils"
+	"go.viam.com/rdk/logging"
+)
+
+func main() {
+	commonMain()
+}
+
+func waitOnline(logger logging.Logger, timeoutCtx context.Context) {
+	logger.Debug("WaitOnline not available on darwin yet")
+}
+
+func ignoredSignal(sig os.Signal) bool {
+	// Ignore SIGURG and other signals that shouldn't cause us to exit
+	return false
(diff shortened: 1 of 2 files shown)


Otterverse (changes requested):
No MAJOR objection to adding this, but if it's going to be non-functional on Darwin, my  gut says "don't add it."

`GOOS=linux make lint` should generally solve the problem you describe. Caveat is that if it builds the linter for the wrong OS, it doesn't run, so, one-line change in the makefile to unset GOOS for that part, and it seems to work fine.

Additionally, this looks like we've never linted windows at all for similar reasons. I'd suggest this change to the Makefile instead:

```
 bin/golangci-lint: Makefile
-       GOBIN=`pwd`/bin go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.64.5
+       GOOS='' GOBIN=`pwd`/bin go install github.com/golangci/golangci-lint/cmd/golangci-lint@v1.64.5
 
 .PHONY: lint
 lint: bin/golangci-lint
        go mod tidy
-       bin/golangci-lint run -v --fix
+       GOOS=linux bin/golangci-lint run -v --fix
+       GOOS=windows bin/golangci-lint run -v --fix

```
That should make it lint both OSes we support. Would probably need to actually clean up the windows linting errors to get that to pass though.

Up to you if you'd rather make the smaller change, or stick with the PR as is. If enabling the darwin build, I'd ask that you add an error message and early return in the main() so people don't accidentally build/run it on Darwin and end up surprised when it doesn't work as a real agent though.

ale7714:
Good point, better to get it working right. Opened a new pr https://github.com/viamrobotics/agent/pull/85 

i don't think it was too bad

PR #84, opened 2025-03-19
[RSDK-10162] Add agent windows installer

ale7714:
* add agent installer powershell script
* add github action to build executable from powershell script
* add make target to build and upload windows installer
* remove wsx since we won't be using msi anymore

https://viam.atlassian.net/browse/RSDK-10162


ale7714:
So the most straightforward way to get an exe is with a powershell script, so i rewrote the bat into powershell. Right now, the installation is pretty verbose, it was helpful for me for testing. I will clean up. 

Also i'm using the MSI workflow for testing but i will probably update the job to make more sense to what's actually happening.

windows-installer-agent.ps1
```diff
@@ -0,0 +1,148 @@
...
+    # Create service
+    $scArgs = @("create", "viam-agent", "binpath=", "`"$agentBinPath`"", "start=", "auto")
+    Start-Process -FilePath "sc" -ArgumentList $scArgs -NoNewWindow -Wait
+    
+    # Configure failure actions
+    $scFailArgs = @("failure", "viam-agent", "reset=", "0", "actions=", "restart/30000/restart/30000/restart/30000")
+    Start-Process -FilePath "sc" -ArgumentList $scFailArgs -NoNewWindow -Wait
+    
+    # Set failure flag
+    $scFlagArgs = @("failureflag", "viam-agent", "1")
```
abe-winter, line 137: not a blocker for this PR, but in the future we'll need to think carefully through the logic here and decide:
- when agent-win should return 0
- how do we 'keep it running at all times' (unless someone asks to shut it down). not straightforward I think bc there is no simple way to request a restart

windows-installer-agent.ps1
```diff
@@ -0,0 +1,148 @@
...
+}
+# Configure and start service
+if (-not $Silent) { Write-Host "Configuring service..." }
+try {
+    # Create service
+    $scArgs = @("create", "viam-agent", "binpath=", "`"$agentBinPath`"", "start=", "auto")
+    Start-Process -FilePath "sc" -ArgumentList $scArgs -NoNewWindow -Wait
+    
+    # Configure failure actions
+    $scFailArgs = @("failure", "viam-agent", "reset=", "0", "actions=", "restart/30000/restart/30000/restart/30000")
```
abe-winter, line 133: is this 30 seconds? should we shorten to 5?

abe-winter (approved):

windows-installer-agent.ps1
ale7714: updated to 5s

windows-installer-agent.ps1
ale7714: good point, i'll think about this next weekk

windows-installer-agent.ps1
```diff
@@ -0,0 +1,148 @@
...
+
+if (-not $Silent) {
+    Write-Host "Starting Viam Agent installation..."
+}
+
+# Define installation paths
+$rootPath = "C:\opt\viam"
+$cachePath = Join-Path $rootPath "cache"
+$binPath = Join-Path $rootPath "bin"
+$agentCURLFileName = "viam-agent-stable-windows-x86_64"
```
abe-winter, line 30: @ale7714 I updated this and the URL below to the stable URL, but have not tested -- kicking this back to you

PR #85, opened 2025-03-20
Update make lint and fix issues

ale7714:


cmd/viam-agent/main.go
```diff
@@ -47,7 +46,6 @@ type agentOpts struct {
 //nolint:gocognit
 func commonMain() {
 	ctx, cancel := setupExitSignalHandling()
```
Otterverse, line 48:
```suggestion
	ctx, globalCancel := setupExitSignalHandling()
```

cmd/viam-agent/main.go
```diff
@@ -47,7 +46,6 @@ type agentOpts struct {
...
-	globalCancel = cancel
 
 	defer func() {
 		cancel()
```
Otterverse, line 51:
```suggestion
		globalCancel()
```

cmd/viam-agent/main.go
```diff
@@ -29,7 +29,6 @@ var (
 
 	// only changed/set at startup, so no mutex.
 	globalLogger = logging.NewLogger("viam-agent")
```
Otterverse, line 31:
```suggestion
	globalLogger = logging.NewLogger("viam-agent")
	globalCancel context.CancelFunc
```

cmd/viam-agent/main.go
```diff
@@ -47,7 +46,6 @@ type agentOpts struct {
 //nolint:gocognit
 func commonMain() {
 	ctx, cancel := setupExitSignalHandling()
-	globalCancel = cancel
```
Otterverse, line 50: This was the only place this was ever set. Without it, the _windows file will always get a nil. I made a few suggestions above/below here which fixes the linter's problem (where it deletes the nolint, then complains about it an unused variable.

cmd/viam-agent/main_linux.go
```diff
@@ -1,3 +1,6 @@
+//go:build linux
+// +build linux
+
```
Otterverse, line 3: Nit: don't use these (rely on the filename itself) OR ALSO add them for the _windows file. I prefer the former, but as long as it's consistent.

cmd/viam-agent/main_windows.go
```diff
@@ -13,8 +13,11 @@ import (
 	"golang.org/x/sys/windows/svc/eventlog"
 )
 
-var elog debug.Log
-var _ svc.Handler = (*agentService)(nil)
+var (
+	elog         debug.Log
+	_            svc.Handler = (*agentService)(nil)
+	globalCancel func()
```
Otterverse, line 19:
```suggestion
```

cmd/viam-agent/main_windows.go
```diff
@@ -41,7 +46,7 @@ func main() {
 	if inService, err := svc.IsWindowsService(); err != nil {
 		panic(err)
 	} else if !inService {
-		println("no service detected -- running as normal process")
+		globalLogger.Info("no service detected -- running as normal process")
```
Otterverse, line 49: Check with @abe-winter on this. "println()" prints to STDERR I believe, not sure if changing this into a log line will prevent it from going into the windows service logs properly or something.

cmd/viam-agent/main_windows.go
```diff
@@ -51,18 +56,24 @@ func main() {
 	if err != nil {
 		return
 	}
-	defer elog.Close()
+	defer func() {
+		//nolint:errcheck,gosec // logging utility errors can be ignored
+		elog.Close()
+	}()
```
Otterverse, line 62:
```suggestion
	defer func() {
		goutils.UncheckedError(elog.Close())
	}()
```
Optional! 

Can use UncheckedError for errors we (generally) don't care about, as it'll still force a log line out in the rare cases an error happens. Same goes for similar lines elsewhere.

utils/utils_windows.go
```diff
@@ -28,7 +28,10 @@ func SyncFS(syncPath string) error {
 	if err != nil {
 		return err
 	}
-	defer syscall.CloseHandle(handle)
+	defer func() {
+		//nolint:errcheck,gosec // CloseHandle errors can be ignored in defer
+		syscall.CloseHandle(handle)
```
Otterverse, line 33:
```suggestion
		goutils.UncheckedError(syscall.CloseHandle(handle))
```
As with similar in above files, Optional!

cmd/viam-agent/main.go
ale7714: tbh i was little confused on how this was working but i get it now

ale7714:
@Otterverse I didnt't know we had `goutils.UncheckedError` but good to know. Replacing all nolint:errcheck with goutils.UncheckedError was a great use for Cursor.

Otterverse (approved):
LGTM!

PR #86, opened 2025-03-20
Update pre-install script to use viam-defaults.json

ale7714:
https://viam.atlassian.net/browse/APP-7902


Otterverse (changes requested):
Kinda minor, but can you change the variable name from PROVISIONING_PATH to "DEFAULTS_PATH" or "VIAM_DEFAULTS_PATH" (or whatever)? My concern is the language around "provisioning" is already very nebulous, and that's now only a small part of what agent (and this file) does.

Otherwise looks good!

Otterverse (approved):

PR #88, opened 2025-03-21
[APP-7889][APP-7791][APP-7636][APP-7946] Bluetooth Provisioning

Otterverse:
Update 03-28: I believe this is now feature complete. I've added full bluetooth functionality to cmd/provisioning-client and validated it works at least linux->linux.

Set the following in the agent version control:
EDIT: Updated to ble.3 rev as of 03-28 afternoon

Added encryption for all client>device writes, using rsa.OAEP, and encoding the public key with x509 PKIX (ASN.1 DER)

```
  "agent": "http://packages.viam.com/temp/viam-agent-v0.16.0-ble.3-aarch64"
```

Outstanding issues:
* ~Automatic pairing: Not working on Android, may not be required at all, need to investigate further. Removed for now.~
* ~Advertised device name may conflict with OS settings for device.~
Update: Both of the above issues were resolved by adding code to disable ReverseDiscovery in the host's bluetooth config. ReverseDiscovery caused the Pi to try to discover services on the client/phone, some of which require higher security (including MIDI audio, which is weirdly the ultimate culprit), and thus initiate a pairing request, which fails, and causes an instant disconnection.

Features:
Can separately disable wifi and ble provisioning in the network_configuration section.
```
  "disable_bt_provisioning": false,
  "disable_wifi_provisioning": false,
```
Note these only block the starting of that provisioning method, not the whole network system like the main "disable_network_configuration" flag in the advanced settings does.

This also fixes a couple of small bugs. Post-upgrade install scripting wasn't running if "wait_for_online..." wasn't set. Similarly, self-disable when there's no wifi or network manager available wasn't working, and would keep retrying and making noise. Both of those _should_ be resolved, but haven't been thoroughly tested yet.

To use the new provisioning client: `go run ./cmd/provisioning-client/` to get started (it'll print basic help/usage)
Add the following options as needed:
* ` --scan` will scan for all (named) bluetooth devices visible.
* `-f SEARCH_TERM` will modify the filter (substring match on the device name.) Optional, and defaults to "viam-setup"
* `-b` enables bluetooth mode
* `-s` shows status
* `-n` shows network list
* `-b --ssid="MY_SSID" --psk="MY_PSK" --appaddr="https://app.viam.com:443" --partID="UUID_PART_ID" --secret="PART_SECRET"` will set both a wifi network AND the machine credentials. Can do only the three for creds or the two for the wifi if not needing/wanting both.

### DIFF
--- a/cmd/provisioning-client/grpc.go
+++ b/cmd/provisioning-client/grpc.go
@@ -1,0 +1,98 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"github.com/viamrobotics/agent/subsystems/networking"
+	pb "go.viam.com/api/provisioning/v1"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials/insecure"
+)
+
+func grpcClient() error {
+	ctx := context.Background()
+
+	conn, err := grpc.NewClient(opts.Address, grpc.WithTransportCredentials(insecure.NewCredentials()))
+	if err != nil {
+		fmt.Println(err)
+	}
+	defer func() {
+		err := conn.Close()
+		if err != nil {
+			fmt.Println(err)
+		}
+	}()
+
+	client := pb.NewProvisioningServiceClient(conn)
+
+	if opts.Status {
+		return GetStatus(ctx, client)
+	}
+
+	if opts.Networks {
+		return GetNetworks(ctx, client)
+	}
+
+	if opts.PartID != "" {
+		return SetDeviceCreds(ctx, client, opts.PartID, opts.Secret, opts.AppAddr)
+	}
+
+	if opts.SSID != "" {
+		return SetWifiCreds(ctx, client, opts.SSID, opts.PSK)
+	}
+	return nil
+}
+
+func GetStatus(ctx context.Context, client pb.ProvisioningServiceClient) error {
+	resp, err := client.GetSmartMachineStatus(ctx, &pb.GetSmartMachineStatusRequest{})
+	if err != nil {
+		return err
+	}
+
+	fmt.Printf("Online: %t, Configured: %t, Provisioning: %v, Last: %v, Errors: %s\n",
+		resp.GetIsOnline(),
+		resp.GetHasSmartMachineCredentials(),
+		resp.GetProvisioningInfo(),
+		resp.GetLatestConnectionAttempt(),
+		strings.Join(resp.GetErrors(), "\n"),
+	)
+	return nil
+}
+
+func GetNetworks(ctx context.Context, client pb.ProvisioningServiceClient) error {
+	resp, err := client.GetNetworkList(ctx, &pb.GetNetworkListRequest{})
+	if err != nil {
+		return err
+	}
+
+	for _, network := range resp.GetNetworks() {
+		fmt.Printf("SSID: %s, Signal: %d%%, Security: %s\n", network.GetSsid(), network.GetSignal(), network.GetSecurity())
+	}
+	return nil
+}
+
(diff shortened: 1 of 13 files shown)


manager.go
```diff
@@ -182,6 +182,10 @@ func (m *Manager) SubsystemUpdates(ctx context.Context) {
 		m.logger.Error(err)
 	}
 	if needRestart {
+		_, err := InstallNewVersion(ctx, m.logger)
+		if err != nil {
+			m.logger.Errorw("running install of new agent version", "error", err)
+		}
```
Otterverse, line 188: This accidentally got removed a couple releases ago, so upgrades weren't running the post-install UNLESS "turn_on_hotspot_if_wifi_has_no_internet" was enabled and it was a fresh restart, which caused it to self-update with slightly different logic  right at startup.

Otterverse:
@maxhorowitz Tagging you here in case you'd like to look at the (hopefully) final version as it got written. No obligation to review, but in case you were curious where it ended up. My apologies, but due to rebasing and splitting PRs, the actual commits you had are buried/lost, so it's just a couple big fat commits credited to me.

FYI, the whole pairing thing you were trying to work around seems to have been a quasi-bug caused by ReverseDiscovery on the device (pi) side. After a moment of connection (or the first write) it'd try to discover services on the client/phone, some of which require higher security/pairing, so it'd request that, but doing so breaks the existing connection. Now I can muddle around in lightblue for as long as I want without getting interrupted (and without pairing) and the go-client also works.

Lastly, and just for posterity, this was the smoking gun when I finally found it: `Mar 27 17:27:31 pluto bluetoothd[1797]: profiles/midi/midi.c:midi_io_initial_read_cb() MIDI I/O: Failed to read initial request`

Yeah, linux/bluez tried to scan for MIDI services on the phone/client (as part of reverse discovery) and that was what tipped me off to it being the cause.

maxhorowitz:
Awesome work James! 

Interesting findings. 

Happy the project got into a good state despite the handoff and OOO interruptions - as always if you want an extra set of hands to test the work I'm more than happy to take a peek - I'm not sure how many sets of hands have "been" in the Agent repo, so I'm more than happy to peek at any work. Thanks for keeping me in the loop.

cmd/provisioning-client/opts.go
Otterverse, line 1: This file is 99% just copied from the original main.go

cmd/provisioning-client/grpc.go
Otterverse, line 1: This file is 99% just copied from the previous main.go

Otterverse:
@abe-winter Discussed with @ale7714 and I think it makes sense to go ahead and merge this, even if the mobile app isn't done yet. Another PR can pick up any fixes from there as needed, but letting this go stale just means other PRs hare more chances to conflict. So, if/when you get a chance if one of you could review the code here, I'd appreciate it. Thanks!

subsystems/networking/bluetooth_linux.go
```diff
@@ -0,0 +1,119 @@
...
+	serviceUUID := bluetooth.NewUUID(uuid.NewSHA1(uuid.MustParse(uuidNamespace), []byte(serviceNameKey)))
+
+	adapter := bluetooth.DefaultAdapter
+	if err := adapter.Enable(); err != nil {
+		return fmt.Errorf("failed to enable bluetooth adapter: %w", err)
+	}
+	if err := adapter.AddService(&bluetooth.Service{UUID: serviceUUID, Characteristics: characteristics}); err != nil {
+		return fmt.Errorf("unable to add bluetooth service to default adapter: %w", err)
+	}
+	// if err := adapter.Enable(); err != nil {
```
ale7714, line 74: remove?

ale7714 (approved):
🚀

PR #89, opened 2025-03-25
[APP-7919] Add option forward Kernel Logs

ale7714:
No rush on getting this merged. This was 95% written by Claude 3.7 and my main goal here was learning how would it go using Cursor Agent.

This PR includes 
* adding forward_kernel_logs (disabled by defaul) config and updating syscfg
* adding test-docker-linux target to make it easier to test changes on linux for non-linux devs. Since I was using Cursor Agent for the code changes, I wanted it to be able to validate the code changes so tests need to be able to run on my environment so cursor can use the output to self correct. 
* adding cursorrules (which i don't need to commit), so you don't have to repeat yourself on every prompt and also to get cursor to produce code in a consistent style. I found this one 👇  to be particularly useful because for some reason it would do unrelated fly-by refactors on every prompt 
> - Only modify code explicitly requested by the change request or task description. Avoid any unrelated or unnecessary changes to maintain clarity and focus 

PS: I also tested this on a Pi and tried enabling, disabling, as well as, restarting viam-agent and it seemed to work fine.

### DIFF
--- a/.cursor/rules/project.mdc
+++ b/.cursor/rules/project.mdc
@@ -1,0 +1,61 @@
+---
+description: 
+globs: 
+alwaysApply: true
+---
+# Viam Agent .cursorrules
+
+The Viam Agent is an open-source service manager that automates the installation, execution, and monitoring of `viam-server` and associated services. It includes a self-update capability to keep itself and the services current, reducing manual maintenance. Additionally, it provides device provisioning and network configuration tools for seamless integration and connectivity within a network.
+
+## Scope of Changes
+- **Relevance**:
+  - Only modify code explicitly requested by the change request or task description. Avoid any unrelated or unnecessary changes to maintain clarity and focus.
+
+## Coding Conventions
+- **Linting Configuration**:
+  - Follow all linting standards as defined in the project's authoritative `golangci.yml` file located at the repository root.
+
+## Architecture & File Structure
+- **Directories**:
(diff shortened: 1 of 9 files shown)


ale7714:
BTW I did ask Claude to use both the agent and rdk as a reference to understand how logging is used and tested. It picked up on the fact that rdk us `logging.NewObservedTestLogger(t)` to inspect log output on tests. I could have probably figure that out but it would have probably taking me longer.

ale7714:
I think the most valuable way to look at this review, it's not to focus on stylistic choices unless they truly go against our standards (and in that case that should be part of the linter) and focus more on the overall design and obviously if something is wrong. 

My prompt for this was "Add a new feature to syscfg that allows users to forward kernel logs, disabled by default, into the existing logging library used by the agent. Parse the kernel log into the log entry format used by rdk/logging." 

I had to do a couple of follow up prompting after because a few things were straight up wrong. But I didn't explicitly tell exactly how to do it and I would like to determine how much detail on the how is necessary and how much are we able to let the ai figure out on its own while still having an acceptable solution.

Otterverse (changes requested):
Mostly a lot of small things, but a few critical ones too.

I know you're busy, and now want to get this out pretty quickly. Would you like me to take a swing at fixing this up quickly today?

ale7714:
Closing this in favor of https://github.com/viamrobotics/agent/pull/89, thanks James!

PR #90, opened 2025-03-25
include windows binary in stable release

abe-winter:
## What changed
- include a `-windows` build in `make all`
- (incidental) don't duplicate transitive deps in upload-stable target; ping me if I misunderstood something
## Why
Windows support


Otterverse (approved):
LGTM

PR #92, opened 2025-03-26
Split OS Building

Otterverse:
Breaking out the bluetooth work into two PRs for easier review. This is splitting the OS builds. Actual changes are very minimal. Renames can't pick up cleanly for a few files because the new generic file uses the old name.

PR #77 is now implemented on top of this for the actual bluetooth work.


abe-winter (approved):
lgtm

~design question, how are the stub subsystems for windows meant to interact with the 'force-disabled' subsystems on main branch?~

~(force disable logic is here https://github.com/viamrobotics/agent/blob/main/utils/config.go#L81-L95)~

~can ask a longer version of this question if helpful~

PR #94, opened 2025-04-01
[RSDK-10343][RSDK-10341][RSDK-10342] Windows Fixes

Otterverse:
General cleanup and bug fixes for windows. Fixes red-herring bugs enumerated in title (mostly all related to something that was accidentally broken by a bad rebase, but actually DO work, so removing the windows-specific bypass logic.) ALSO fixes the use of "file:///" version control URLs on windows.

### DIFF
--- a/utils/utils.go
+++ b/utils/utils.go
@@ -57,6 +57,7 @@ func GetRevision() string {
 
 func init() {
 	if runtime.GOOS == "windows" {
+		ViamDirs["viam"] = "c:/opt/viam"
 		// note: forward slash isn't an abs path on windows, but resolves to one.
 		var err error
 		ViamDirs["viam"], err = filepath.Abs(ViamDirs["viam"])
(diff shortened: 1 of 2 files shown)


utils/utils.go
```diff
@@ -53,6 +53,7 @@ func GetRevision() string {
 
 func init() {
 	if runtime.GOOS == "windows" {
+		ViamDirs["viam"] = "c:/opt/viam"
```
ale7714, line 56: Is it always c:? I thought we had concerns about this

ale7714 (approved):
One question otherwise makes sense I think

utils/utils.go
Otterverse: No, it's probably not... but previously we were just going with "random drive" and not even specifying it, so that was worse.

Likely needs to look for some ENV var to get the WINDOWS_PATH or something, and determine the system drive from there. Had meant to pursue further.

Otterverse: One further thought... this is correct for at least as long as we're hardcoding that same path/drive in the installer (which it looks like we are.) When we fix this, we really should be putting it in %PROGRAM_FILES% or %APPLICATION_DATA% and stuff, not unix-style paths like "/opt" anyway.

PR #96, opened 2025-04-08
[APP-7946] Additional Bluetooth Provisioning Changes

Otterverse:
Further work to support the bluetooth provisioning method and integrate with the bluetooth socks proxy work.


cmd/provisioning-client/bluetooth.go
```diff
@@ -30,6 +30,9 @@ const (
 	appAddressKey            = "app_address"
 	availableWiFiNetworksKey = "networks"
 	statusKey                = "status"
+	manufacturerKey          = "manufacturer"
+	modelKey                 = "model"
+	fragmentKey              = "fragment_id"
```
kevin49999, line 35: Making a change on my end to read the fragment id, should i read the manufacturer and model for any reason? The only thing I could think of was for displaying it in the UI in some way (from my perspective)

kevin49999: Think I got an answer here! https://github.com/viamrobotics/canyon-runner-mobile/pull/190#discussion_r2043267806

preinstall.sh
```diff
@@ -13,6 +13,7 @@ SERVICE_FILE=$(cat <<EOF
 [Unit]
 Description=Viam Services Agent
 After=NetworkManager.service
+After=bluetooth.service
```
abe-winter, line 16: thinking through consequences of the After

[systemd docs for After directive](https://www.freedesktop.org/software/systemd/man/latest/systemd.unit.html#Before=)

it sounds like this is only respected if 'both units are being started' (i.e. systems without bluetooth will still be able to start agent)

note this has potential to delay startup on systems which have bluetooth -- this is not a merge blocker, but consider pinging SEs before actually releasing this

abe-winter (commented):
did read-through to look for non-provisioning + windows issues -- one comment about startup time, no critical flags

cmd/provisioning-client/bluetooth.go
Otterverse: These are purely informative for generic whitelabeling purposes. Safe to ignore, but I wanted to match parity with with web/grpc provisioning APIs.

preinstall.sh
Otterverse: Yes, it used to be "Wants=NetworkManager" and that caused problems for systems without. After means "if you're starting it anyway, sequence."

I did a couple timing tests, bluetooth starts up in basically the same second as NetworkManager on my Pi (same timestamp) so I don't think the delay is an issue. And it's definitely an issue if bluetooth provisioning fails to work because it started up in the wrong order. Will definitely let SE know anyway.

subsystems/networking/bluetooth_linux.go
```diff
@@ -89,28 +107,96 @@ func (n *Networking) initializeBluetoothService(deviceName string, characteristi
...
+		return errw.Wrap(err, "exporting custom agent object")
 	}
 
+	obj := conn.Object(BluezDBusService, "/org/bluez")
+	call := obj.Call("org.bluez.AgentManager1.RegisterAgent", 0, dbus.ObjectPath(BluezAgentPath), "NoInputNoOutput")
+	if err := call.Err; err != nil {
+		return errw.Wrap(err, "registering custom agent")
+	}
+
+	call = obj.Call("org.bluez.AgentManager1.RequestDefaultAgent", 0, dbus.ObjectPath(BluezAgentPath))
```
benjirewis, line 156: [q] Any reason to make your `NoInputNoOutput` agent the _default_ agent on the device? Were you having issues without this?

benjirewis (approved):
From my perspective looks great!

subsystems/networking/bluetooth_linux.go
Otterverse: The normal "register agent" functions in other libraries all do this. The idea was maybe I could hook the incoming requests from the pop-up. Otherwise, the registered agent is supposed to ONLY deal with pairing requests our app initiates (and we don't initiate pairing requests ourselves.) Feel free to play around with it.

PR #103, opened 2025-04-24
Allow provisioning mode to exit quickly when new wifi is added, fix bluetooth tethering retry rate.

Otterverse:
This fixes two subtle but ugly bugs. The first was that provisioning wouldn't always exit right away when new wifi was added. The second is that if a bluetooth tethering connection is present (but not working, e.g. disabled on phone that's paired) limit the connection attempts to the visibility timeout (one minute.)

Test build: https://storage.googleapis.com/packages.viam.com/temp/viam-agent-v0.18.0-rc1-aarch64


ale7714 (approved):
good catch 🕵️

PR #105, opened 2025-04-30
Reduce most error logging to warning level

Otterverse:
Reducing most Error() calls to Warn() equivalents


subsystems/viamserver/viamserver.go
```diff
@@ -93,7 +93,7 @@ func (s *viamServer) Start(ctx context.Context) error {
 	}
 
 	stdio := utils.NewMatchingLogger(s.logger, false, false)
-	stderr := utils.NewMatchingLogger(s.logger, true, false)
+	stderr := utils.NewMatchingLogger(s.logger, false, false)
```
ale7714, line 96: is this because we want stderr to go out as info? sorry i can't tell what the level will actually be

Otterverse: That's the "isError" argument, and means "default to error when log level is unknown" (for the "unstructcured output" stuff) and I don't think we want that anymore. Viam seems to always spit out some unstructured output during startup.

ale7714: ohhh yeah that make sense

ale7714 (approved):

PR #106, opened 2025-05-01
Fix updates to use already-downloaded files

Otterverse:
This fixes a bug introduced a few versions back that causes re-downloads to occur even if a version of a binary was already downloaded/cached.

### DIFF
--- a/version_control.go
+++ b/version_control.go
@@ -184,7 +184,7 @@ func (c *VersionCache) UpdateBinary(ctx context.Context, binary string) (bool, e
 		return false, errw.Errorf("unknown binary name for update request: %s", binary)
 	}
 
-	var needRestart bool
+	var needRestart, goodBytes bool
 
 	if data.brokenTarget {
 		return needRestart, nil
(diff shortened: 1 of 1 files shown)


version_control.go
```diff
@@ -233,50 +235,53 @@ func (c *VersionCache) UpdateBinary(ctx context.Context, binary string) (bool, e
 	// this is a new version
 	c.logger.Infof("new version (%s) found for %s", verData.Version, binary)
 
-	// download and record the sha of the download itself
-	var err error
-	verData.DlPath, err = utils.DownloadFile(ctx, verData.URL, c.logger)
-	if err != nil {
-		if isCustomURL {
-			data.brokenTarget = true
+	if !goodBytes {
```
Otterverse, line 238: The bulk of this fix is just moving a lot of stuff (that should only happen when a redownload IS needed) in this block, and re-ordering one or two other snippets. Nearly zero "new" code.

abe-winter (approved):
lgtm

there's some test suite coverage here + I'm guessing you tested manually as well, but would not mind 1) splitting up and 2) adding coverage for UpdateBinary in the future. it does a lot and there are many paths through this function

Otterverse:
> lgtm
> 
> there's some test suite coverage here + I'm guessing you tested manually as well, but would not mind 1) splitting up and 2) adding coverage for UpdateBinary in the future. it does a lot and there are many paths through this function

Agree this should get broken up and tested. This was just trying to get it back to "somewhat working" for now, between higher priority tasks.

PR #107, opened 2025-05-02
Set debug from viam-defaults.json early

Otterverse:
This allows debug (if set in viam-defaults.json) to be applied early, before provisioning starts, rather than only after "normal" operations have started.


manager.go
```diff
@@ -276,14 +276,6 @@ func (m *Manager) CheckUpdates(ctx context.Context) time.Duration {
 		interval = minimalCheckInterval
 	}
 
-	m.cfgMu.RLock()
```
ale7714, line 279: is the reason locking is not needed anymore is that is happening before we enter the loop where the config could change?

ale7714 (approved):

PR #108, opened 2025-05-05
[APP-8059] [APP-7835] Improve windows service functionality

Otterverse:
This fixes the startup and health checks for viam server by no longer writing to the (non-existant) STDOUT when in windows service mode. It also allocates a console, allowing interrupt signaling for graceful shutdown of viam-server. Lastly, it improves handling/adding the ".exe" suffix for windows. Appending it immediately upon reception from the cloud, rather than manipulating it multiple places later.

### DIFF
--- a/cmd/viam-agent/main_windows.go
+++ b/cmd/viam-agent/main_windows.go
@@ -4,6 +4,9 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"os/exec"
+	"strconv"
+	"strings"
 
 	"github.com/viamrobotics/agent"
 	"github.com/viamrobotics/agent/utils"
@@ -80,8 +83,7 @@ func main() {
 	}
 	// wait first so viam-server doesn't try to restart
 	activeBackgroundWorkers.Wait()
-	// KillTree to catch any stragglers
-	if err := utils.KillTree(-1); err != nil {
+	if err := zapChildren(); err != nil {
 		goutils.UncheckedError(elog.Error(1, fmt.Sprintf("error killing subtree %s", err)))
 	}
 	goutils.UncheckedError(elog.Info(1, fmt.Sprintf("%s service stopped", serviceName)))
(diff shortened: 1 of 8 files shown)


Otterverse:
@abe-winter I think this MAY also resolve https://viam.atlassian.net/browse/APP-8165 as I don't see the symptoms reported with this fix in place. I'm not 100% certain I understand the issue there though, so please try/confirm yourself if you can. Thanks!

Otterverse:
> @abe-winter I think this MAY also resolve https://viam.atlassian.net/browse/APP-8165 as I don't see the symptoms reported with this fix in place. I'm not 100% certain I understand the issue there though, so please try/confirm yourself if you can. Thanks!

In hindsight, this probably doesn't cover that. I think I misunderstood the description in 8165. This was about weird errors showing up for modules on shutdown. I somehow conflated it with the orphan processes issue.

abe-winter:
testing
- [x] healthcheck URLs work in service mode
> 5/7/2025, 7:57:18 PM info viam-agent   viamserver/viamserver.go:153   healthcheck URLs: https://MACHINE.LOCATION.local.viam.cloud:8080 https://localhost:8080 

- [x] process control: toggled back and forth between 0.74.0 and 0.73.0, had one copy of module process after each change

cmd/viam-agent/main_windows.go
```diff
@@ -49,6 +53,11 @@ func main() {
 		return
 	}
 
+	// in service mode we have to alloc our own console to be able to send interrupts
+	if r, _, err := windows.NewLazySystemDLL("kernel32.dll").NewProc("AllocConsole").Call(); r == 0 {
```
abe-winter, line 57: skimming docs for AllocConsole https://learn.microsoft.com/en-us/windows/console/allocconsole

questions:
1. 'the AllocConsole function fails if the calling process already has a console' -- guessing this is why you only do this in service mode
2. are there circumstances under which this creates a floating cmd.exe box? guessing not bc we haven't seen one
3. do you need to call freeconsole somewhere so this doesn't leak? is there a system limit on how many of these can be created?

utils/utils.go
```diff
@@ -111,13 +111,11 @@ func DownloadFile(ctx context.Context, rawURL string, logger logging.Logger) (ou
...
-
 	parsedPath := parsedURL.Path
-	if runtime.GOOS == "windows" && !strings.HasSuffix(parsedPath, ".exe") {
-		parsedPath += ".exe"
-	}
-
 	outPath = filepath.Join(ViamDirs["cache"], path.Base(parsedPath))
+	if runtime.GOOS == "windows" && !strings.HasSuffix(outPath, ".exe") {
+		outPath += ".exe"
+	}
```
abe-winter, line 118: why move this down so it applies to `outPath` but not `parsedPath`?

utils/utils_windows.go
```diff
@@ -1,22 +1,34 @@
...
 	goutils "go.viam.com/utils"
-	"golang.org/x/sys/windows/svc/eventlog"
+	"golang.org/x/sys/windows"
+	"golang.org/x/sys/windows/svc"
 )
 
-func PlatformProcSettings(cmd *exec.Cmd) {}
+func PlatformProcSettings(cmd *exec.Cmd) {
+	cmd.SysProcAttr = &syscall.SysProcAttr{
+		CreationFlags: syscall.CREATE_NEW_PROCESS_GROUP,
```
abe-winter, line 18: what does CREATE_NEW_PROCESS_GROUP do here? permit console inheritance?

abe-winter (approved):
lgtm, left pedantic questions

utils/utils.go
Otterverse: Because of the `file://` handler below. We'll use the original parsedPath to open and copy the source file. If we appened .exe to it, now it's an invalid filename that doesn't exist. (Hit this in windows testing.)

PR #109, opened 2025-05-06
[APP-7500] [RSDK-10333] Automatically restart agent when update available (and safe)

Otterverse:
Note: This should merge after #108 and until then, will look like a larger set of changes.

This update makes viam-agent exit gracefully after it has installed a new update for itself (and optionally viam-server) when it's safe to do so. That is, when viam-server is disabled, or reports "safe to restart."

This also fixes a bug on windows preventing lockups or "self-exit" there from working as intended.

### DIFF
--- a/cmd/viam-agent/main_windows.go
+++ b/cmd/viam-agent/main_windows.go
@@ -35,12 +35,13 @@ func (*agentService) Execute(args []string, r <-chan svc.ChangeRequest, changes 
 		c := <-r
 		if c.Cmd == svc.Stop || c.Cmd == svc.Shutdown {
 			goutils.UncheckedError(elog.Info(1, fmt.Sprintf("%s service stopping", serviceName)))
+			changes <- svc.Status{State: svc.StopPending}
+			globalCancel()
 			break
 		} else {
 			goutils.UncheckedError(elog.Error(1, fmt.Sprintf("unexpected control request #%d", c)))
 		}
 	}
-	changes <- svc.Status{State: svc.StopPending}
 	return
 }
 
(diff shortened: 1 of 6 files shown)


cmd/viam-agent/main_windows.go
```diff
@@ -59,23 +69,16 @@ func main() {
...
+
+	go func() {
+		// note: svc.Run() hangs until windows terminates the service via Execute()
+		if err := svc.Run(serviceName, &agentService{}); err != nil {
+			goutils.UncheckedError(elog.Error(1, fmt.Sprintf("%s service failed: %v", serviceName, err)))
+		}
+	}()
+
+	commonMain()
+	if err := zapChildren(); err != nil {
```
Otterverse, line 81: I don't know if this is technically needed or not. It was the original utils.KillTree (but it doesn't kill the tree, just the leaves) and seems to only be to try to collect/kill orphans, which should be cleaned up by viam-server itself now that we're doing things gracefully. But... extra clean-up didn't seem to do any harm, and if it helps fence in bad modules code, why not.

Also,  I ALMOST renamed it OrphanCrushingMachine() :-)

subsystems/viamserver/viamserver.go
```diff
@@ -137,7 +137,7 @@ func (s *viamServer) Start(ctx context.Context) error {
 		if s.cmd.ProcessState != nil {
 			s.lastExit = s.cmd.ProcessState.ExitCode()
 			if s.lastExit != 0 {
-				s.logger.Errorw("non-zero exit code", "exit code", s.lastExit)
+				s.logger.Error("non-zero exit code: %d", s.lastExit)
```
Otterverse, line 140: Needs to be Errorf() not Error() (will correct after 108 is in and can more easily rebase)

abe-winter:
I changed the base branch from `main` to `windowsagogo` to make this easier to read -- will change back

cmd/viam-agent/main_windows.go
```diff
@@ -123,7 +118,8 @@ func zapChildren() error {
 		goutils.UncheckedError(elog.Info(1, fmt.Sprintf("KillTree stopping %d children of pid %d", len(lines), pid)))
 	}
 	for _, line := range lines[1:] {
-		if line == "" {
+		line = strings.TrimSpace(line)
+		if _, err := strconv.Atoi(line); err != nil {
```
abe-winter, line 122: are you parsing the pid twice now?

manager.go
```diff
@@ -195,32 +199,39 @@ func (m *Manager) SubsystemUpdates(ctx context.Context) {
...
 
-		if needRestart || m.viamServerNeedsRestart {
+		if needRestart || m.viamServerNeedsRestart || m.viamAgentNeedsRestart {
 			if m.viamServer.(viamserver.RestartCheck).SafeToRestart(ctx) {
 				if err := m.viamServer.Stop(ctx); err != nil {
 					m.logger.Warn(err)
 				} else {
 					m.viamServerNeedsRestart = false
 				}
+				if m.viamAgentNeedsRestart {
```
abe-winter, line 231: in cases where both agent + rdk need restart, does it make sense to check agent first, so we can restart once instead of twice? or am I thinking about this wrong

abe-winter (approved):
guessing we'll test some restart cases after merging, but lgtm

cmd/viam-agent/main_windows.go
Otterverse: This is more just a test that we're getting JUST an integer from the line (as opposed to doing a regex or something on the line) as I was getting a lot of errors in the event log from the "not a valid childProcess line" below. Could be more elegant. But for something that fires only once each run, I didn't overthink it.

manager.go
Otterverse: That's what this is... we already SET this value above for agent. We're in the block checking viam-server now. So if viam server says it's safe to restart, then it doesn't matter if it NEEDS a restart, it's getting one anyway. But we're past the point where it too, would've been updated. So at this point, possibly BOTH have had new binaries installed, and agent (at least) needs to restart, so we do it once for both of them.

PR #110, opened 2025-05-06
[RSDK-9293] Automatically prune cached binaries

Otterverse:
Note: Should merge after #109 as this inherits those changes and looks like bigger change set until then

This adds a CleanCache() function which prunes the cache and removes any obsolete/stray files to recover disk space. It runs every 24 hours, but won't run within the first 15 minutes of startup to avoid adding load on machines that may have been offline for an extended period and have a lot of updates to get through.

### DIFF
--- a/version_control.go
+++ b/version_control.go
@@ -12,6 +12,7 @@ import (
 	"path"
 	"path/filepath"
 	"runtime"
+	"slices"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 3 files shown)


Otterverse:
Found some small bugs in manual testing. Working through those. Will re-request review after re-vetting.

version_control.go
```diff
@@ -307,12 +317,90 @@ func (c *VersionCache) UpdateBinary(ctx context.Context, binary string) (bool, e
...
-func mimeIsAny(mtype *mimetype.MIME, expected []string) bool {
-	for _, expectedType := range expected {
-		if mtype.Is(expectedType) {
-			return true
+func (c *VersionCache) CleanCache(ctx context.Context) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	// this can be set to the number of days to keep, ex: "VIAM_AGENT_FORCE_CLEAN=5"
+	forceVal := os.Getenv("VIAM_AGENT_FORCE_CLEAN")
```
abe-winter, line 325: optional: put in agentOpts here https://github.com/viamrobotics/agent/blob/main/cmd/viam-agent/main.go#L36, will get docstring, default + int conversion for free (I assume -- have never used our parser library for ints)

abe-winter (approved):

version_control.go
Otterverse: Didn't do that as this is almost certainly ONLY ever for testing. Secondly, I'd have to then pass the opts struct all the way down the stack to here. That didn't seem ideal for a "test-only" thing.

PR #114, opened 2025-05-15
Skip healthchecks on windows

Otterverse:
Hard code return values for windows to workaround issue with windows suspend/resume.


abe-winter (approved):

ale7714:
please, let's publish a patch with this and we can test separately healtchecks and fixes

PR #116, opened 2025-05-20
RSDK-10704, RSDK-10292: Config: new setting viam_server_env to pass environment variables to viam-server

aldenh-viam:
The `advanced_settings` section of the config now supports `viam_server_env` to pass user-specified environment variables to `viam-server` and its children.

Example:
```
. . . 
"advanced_settings": {
    "viam_server_env": {
      "MODULE_DEBUG": "1",
      "BLUETOOTH": "FALSE",
    }
}
. . . 
```

Keys and values must both be strings. To unset, remove from the mapping.

Note all of the existing environment variables from Agent's current environment currently also gets passed to `viam-server`, e.g. HOME, PWD, TERM, PATH, etc.

### DIFF
--- a/examples/agent-config.jsonc
+++ b/examples/agent-config.jsonc
@@ -12,7 +12,10 @@
 		"viam_server_start_timeout_minutes": 10,
 		"disable_viam_server": false,
 		"disable_network_configuration": false,
-		"disable_system_configuration": false
+		"disable_system_configuration": false,
+		"viam_server_env": {
+			"MODULE1_DEBUG": "1"
+		}
 	},
 	"network_configuration": {
 		"manufacturer": "viam",


subsystems/viamserver/viamserver.go
```diff
@@ -102,6 +105,15 @@ func (s *viamServer) Start(ctx context.Context) error {
 	s.cmd.Stdout = stdio
 	s.cmd.Stderr = stderr
 
+	if len(s.extraEnvVars) > 0 {
+		// if s.cmd.Env is not explicitly specified (nil), viam-server would inherit all env vars in Agent's environment
+		s.cmd.Env = s.cmd.Environ()
+		for k, v := range s.extraEnvVars {
+			s.cmd.Env = append(s.cmd.Env, k+"="+v)
+		}
+		s.logger.Debugf("Starting viam-server with environment variables %v", s.cmd.Env)
```
cheukt, line 114: let's add an Info log for `s.extraEnvVars` and we can keep the debug log for the full env

cheukt (commented):
does the viam-server restart if you update env vars on app.viam.com?

manager.go
```diff
@@ -216,6 +218,12 @@ func (m *Manager) SubsystemUpdates(ctx context.Context) {
 		if err != nil {
 			m.logger.Warn(err)
 		}
+		if !reflect.DeepEqual(m.cfg.AdvancedSettings.ViamServerExtraEnvVars, m.prevCfg.AdvancedSettings.ViamServerExtraEnvVars) {
+			m.logger.Infow("Detected changed environment variables. Restarting viam-server at next opportunity.",
+				"current", m.cfg.AdvancedSettings.ViamServerExtraEnvVars,
+				"previous", m.prevCfg.AdvancedSettings.ViamServerExtraEnvVars)
+			needRestart = true
+		}
```
cheukt, line 226: somewhat minor, but is there a reason to have the needs restart logic here instead of inside the viamserver.Update method?

utils/config.go
```diff
@@ -28,6 +28,7 @@ var (
 			DisableNetworkConfiguration:   Tribool(0),
 			DisableSystemConfiguration:    Tribool(0),
 			ViamServerStartTimeoutMinutes: Timeout(time.Minute * 10),
+			ViamServerExtraEnvVars:        nil,
```
Otterverse, line 31: Please make sure to add this (and in the future, any other new config fields) to examples/agent-config.jsonc

cheukt (approved):
LGTM!

utils/config.go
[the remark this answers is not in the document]
Otterverse: You don't NEED it, but you don't need a lot of the other zero-values in this block either. They're here to make code more understandable, and to guard against any accidental change of type(s) in the future. E.g. it's easier to look at a "complete" structure here than see only the non-zero fields. 

tl;dr I'd prefer it stay.

PR #118, opened 2025-06-03
Add enough support for macOS for compilation and unit tests to work

jmatth:


utils/utils_darwin.go
```diff
@@ -0,0 +1,9 @@
+package utils
+
+import (
+	"golang.org/x/sys/unix"
+)
+
+func syncfs(fd uintptr) error {
+	return unix.Fsync(int(fd))
```
Otterverse, line 8: This isn't the same as Syncfs, this only syncs one file, not everything on the filesystem.

Otterverse (changes requested):
What's the scope/reason for these changes? Previously, it was INTENDED that agent not support darwin directly, so if people accidentally build or test there natively there it would fail and not accidentally result in a binary that doesn't run where intended, or the illusion that tests are passing (since all the linux only code wouldn't get tested.)

As for making tests work so you don't have to rely on CI... that's the purpose of canon. So one can simply run `canon make test` or `canon make all` and it "just works."

If you're not familiar with canon, my apologies (we should probably add it to the readme!) Details here: https://github.com/viamrobotics/canon

jmatth:
The motivation is purely to allow for development/testing of the go-only parts of the code on macos. I had previously attempted to use cannon and after some fiddling it's what finally pushed my system into asking which process I would like to be oom killed, so it would be nice to not have to rely on it when not working on the linux parts of the codebase.

> ...if people accidentally build or test there natively there it would fail and not accidentally result in a binary that doesn't run where intended...

Good point, I hadn't considered someone external w/o the full context trying to produce a macos build. Easiest fix while still making these changes is probably to change the main function in `main_unsupported.go` to report the platform is unsupported and then exit. Then we still get a minimal working dev environment on macos + nobody can get a working macos build without modifying the source code.

jmatth:
For a) I mostly agree and the problem was probably more caused by already running Slack (Chrome), VSCode (Chrome), Firefox, gopls, etc. At the same time, Docker Desktop is way more resource hungry on non-Linux machines because it has to run a VM to host the containers. For an extremely unscientific experiment I just shut down Docker on my system after using it to run several canon commands and instantly reclaimed 5GB of memory.

> What's the point of adding stuff to build for MacOS when you're talking about working on and testing the LINUX code?

I'm not talking about the Linux code, I'm talking about the cross-platform parts of the code.

- Yes `github.com/viamrobotics/agent/subsystems/syscfg	[no test files]`, but also `ok      github.com/viamrobotics/agent   0.808s` and `ok      github.com/viamrobotics/agent/utils     0.949s`. 68% of the existing tests work fine on macos.
- One of the past PRs I was sent as an example of what I should look into adding tests for was #73. That PR includes a bunch of NetworkManager changes but the fix for the cached config behavior is not tied to a specific platform and could have a test written that runs anywhere go does.

> Is there some specific point of friction in your development flow that maybe I'm missing or not understanding?

Not that prompted this PR.

Otterverse (approved):
As I mentioned in Slack, my intent here was not to be stubborn on this, so sorry for the back and forth. I still don't fully understand how this reduces friction for Mac development, but maybe that's just preference or workflow difference on my end as I develop infrequently on Mac. And as we are far from having complete tests in any case (and they are tested in GH workflows now.) So if this PR is useful to YOU, then good enough.

PR #119, opened 2025-06-03
[APP-8431] Bluetooth Improvments

Otterverse:
This adds a new “ExitProvisioning()” grpc call, and “agent_version” field to MachineStatus(), along with equivalent characteristics for bluetooth. The new behavior is that provisioning clients (mobile app) must call ExitProvisioning() after they are done (instead of relying on heuristics like time-since-last-write and/or “guessing” when a group of writes is complete), causing the device to exit provisioning mode immediately, and attempt to resume normal operations.

This also adds a pre-shared key to the written characteristics. This PSK is the same as the hotspot password, and should be prepended to written charcteristics and separated with a colon, before encryption. Ex:

SSID = "myHotspotPassword:" + SSID

This is simple fix to close the security hole where a malicious actor could potentially claim another user's device (and thus gain access) by causing it to go offline (ex: through wifi jamming) and replacing the credentials with their own. A user can change/override the hotspot password from the cloud once they first activate a device to make it unique to them (instead of it always using the manufacturer set value in viam-defaults.json) when concerned about this.

### DIFF
--- a/subsystems/networking/bluetooth_characteristics_linux.go
+++ b/subsystems/networking/bluetooth_characteristics_linux.go
@@ -8,6 +8,7 @@ import (
 	"crypto/x509"
 	"errors"
 	"fmt"
+	"strings"
 	"sync"
 
 	"github.com/google/uuid"
@@ -56,13 +57,15 @@ type btCharacteristics struct {
 	userInputData *userInputData
 
 	privKey *rsa.PrivateKey
+	PSK     string
 }
 
-func newBTCharacteristics(logger logging.Logger, userInputData *userInputData) *btCharacteristics {
+func newBTCharacteristics(logger logging.Logger, userInputData *userInputData, psk string) *btCharacteristics {
 	return &btCharacteristics{
 		logger:        logger,
 		writables:     map[string]*bluetooth.Characteristic{},
 		userInputData: userInputData,
+		PSK: psk,
 	}
 }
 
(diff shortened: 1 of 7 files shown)


cheukt:
agree with the changes for BT, but why do we need the ExitProvisioning() grpc call? mostly concerned about the breaking change this introduces and how it'll either force users to need to upgrade their mobile app even if BT is unneeded/unwanted or force us to permanently pin agent to before this change for existing users.

we know for certain whether a full set of wifi creds or full set of viam creds are sent because of the associated grpc calls, what does ExitProvisioning() provide that we can't infer from those two?

jckras:
would we need to update all current provisioning flows to use exitProvisioning()? I am not sure I understand how different versions of agent work when we already have devices out there in the wild? Similar to @cheukt's question but would this be a breaking change or would our mobile apps still work with the current provisioning flow?

Otterverse:
> agree with the changes for BT, but why do we need the ExitProvisioning() grpc call? mostly concerned about the breaking change this introduces and how it'll either force users to need to upgrade their mobile app even if BT is unneeded/unwanted or force us to permanently pin agent to before this change for existing users.

That's true for any breaking changes we ever make to this API though, and pinning versions is supposed to be standard practice for that reason, at least with our known clients. On top of that, we're also going to be adding back tethering and (on some hardware) a "pair" button to enter provisioning mode manually. In those scenarios it's even harder to know when a user is "done" with their tasks. 

> 
> we know for certain whether a full set of wifi creds or full set of viam creds are sent because of the associated grpc calls, what does ExitProvisioning() provide that we can't infer from those two?

But are you sending both? Just one? Which one? Are you going to pair (in the phone's OS) as part of tethering or are you done with that too? ExitProvisioning() provides a way to say "I'm done now, forget heuristics"  and keeps parity between bluetooth and wifi methods.



> would we need to update all current provisioning flows to use exitProvisioning()? I am not sure I understand how different versions of agent work when we already have devices out there in the wild? Similar to @cheukt's question but would this be a breaking change or would our mobile apps still work with the current provisioning flow?


The idea is that one would check the agent version (now a field in machine status) and would know if that's newer than X .Y then ExitProvisioning() would be expected. Otherwise the old flow can continue to work. My understanding is that clients (with mobile provisioning) have pinned versions of agent so they can remain on their current version for as long as needed until mobile changes are implemented.

cheukt:
that makes sense, thanks for the explanation!

PR #121, opened 2025-06-20
RSDK-10869 - Always send inputs from wifi provisioning

cheukt:
fully aware this will be overwritten by https://github.com/viamrobotics/agent/pull/119, but this is a temp fix


cheukt:
still testing but promising so far

PR #122, opened 2025-06-24
RSDK-10985 - Remove viam-server healthcheck

cheukt:
should remove restarts on false positives and stop hiding issues with the viam-server if it is indeed unresponsive.


dgottlieb (approved):
Can't tell if other bits weren't changed that ought to be changed. But this looks good to me.

Otterverse (approved):
LGTM

PR #123, opened 2025-06-24
RSDK-11021 - Only log ssid truncated if it is user defined

cheukt:
side steps the whole only validate config if it is different from previous, but didn't think we needed to do that yet

the idea is that we would only warn if this is a deliberate choice that the user made, but also down to just remove the log entirely

### DIFF
--- a/utils/config.go
+++ b/utils/config.go
@@ -392,6 +392,10 @@ func validateConfig(cfg AgentConfig) (AgentConfig, error) {
 			errw.New("network_configuration.hotspot_password should not be empty, please omit empty fields entirely"))
 	}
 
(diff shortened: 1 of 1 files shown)


Otterverse (approved):
I suppose this works for now as a simpler fix than diffing the configs.

cheukt:
> Also, why does moving this up a block help?

the current behavior can be confusing because this can happen without users even doing anything - everything can be on default but if your hostname is too long (which happens on a lot of windows/personal computers), this warning will show up every 5s. 

the change makes it so that this warning only spews if the user deliberately sets the hotspot ssid in the config to be too long, which might not be the most complete solution (I agree that that diffing and only validating when there is a change is likely more complete), but at least the user would've had some input.

PR #124, opened 2025-06-25
RSDK-10951: Quickfix for more responsive UX on incorrect wifi password

aldenh-viam:
Quickfix to bypass the internal hotspot restart timer in specific cases (namely, when user enters incorrect wifi password). Works well on my Raspberry Pi 4B with both the web portal and GRPC provisioning, but let me know if there are other odd devices or corner cases. I also invite others to help test.

Switching between networks on the OS side (disabling hotspot in order to try the newly entered wifi, and reactivating hotspot if it fails) takes some time, so this is likely about the fastest and most "responsive" I can see with this current design.

```
2025-06-25T20:33:14.834Z        INFO    viam-agent      networking/networkmanager_linux.go:204  Starting provisioning mode.
2025-06-25T20:33:14.835Z        INFO    viam-agent      networking/networkmanager_linux.go:501  Adding/updating settings for network viam-setup-10951@wlan0
2025-06-25T20:33:14.853Z        INFO    viam-agent      networking/networkmanager_linux.go:306  Activating connection: viam-setup-10951@wlan0
2025-06-25T20:33:18.861Z        INFO    viam-agent      networking/networkmanager_linux.go:356  Successfully activated connection: viam-setup-10951@wlan0
2025-06-25T20:33:18.861Z        INFO    viam-agent      networking/networkmanager_linux.go:247  Hotspot provisioning set up successfully.
2025-06-25T20:33:18.942Z        INFO    viam-agent      networking/bluetooth_linux.go:64        Bluetooth provisioning started.
(waiting for user input...)
2025-06-25T20:33:44.705Z        INFO    viam-agent      networking/networkmanager_linux.go:744  Wifi settings received for Viam-2G
2025-06-25T20:33:44.748Z        INFO    viam-agent      networking/networkmanager_linux.go:501  Adding/updating settings for network Viam-2G@wlan0
2025-06-25T20:33:46.786Z        INFO    viam-agent      networking/networkmanager_linux.go:258  Stopping provisioning mode.
2025-06-25T20:33:46.787Z        INFO    viam-agent      networking/networkmanager_linux.go:390  Deactivating connection: viam-setup-10951@wlan0
2025-06-25T20:33:46.809Z        INFO    viam-agent      networking/networkmanager_linux.go:397  Successfully deactivated connection: viam-setup-10951@wlan0
2025-06-25T20:33:46.809Z        INFO    viam-agent      networking/networkmanager_linux.go:280  Stopped hotspot provisioning mode.
-->2025-06-25T20:33:53.196Z        INFO    viam-agent      networking/networkmanager_linux.go:306  Activating connection: Viam-2G@wlan0
-->2025-06-25T20:34:06.213Z        WARN    viam-agent      networking/networkmanager_linux.go:620  bad or missing password
-->2025-06-25T20:34:06.213Z        INFO    viam-agent      networking/networkmanager_linux.go:204  Starting provisioning mode.
2025-06-25T20:34:06.213Z        INFO    viam-agent      networking/networkmanager_linux.go:501  Adding/updating settings for network viam-setup-10951@wlan0
2025-06-25T20:34:06.232Z        INFO    viam-agent      networking/networkmanager_linux.go:306  Activating connection: viam-setup-10951@wlan0
2025-06-25T20:34:10.246Z        INFO    viam-agent      networking/networkmanager_linux.go:356  Successfully activated connection: viam-setup-10951@wlan0
2025-06-25T20:34:10.247Z        INFO    viam-agent      networking/networkmanager_linux.go:247  Hotspot provisioning set up successfully.
2025-06-25T20:34:10.318Z        INFO    viam-agent      networking/bluetooth_linux.go:64        Bluetooth provisioning started.
(waiting for user input...)
2025-06-25T20:34:31.224Z        INFO    viam-agent      networking/networkmanager_linux.go:744  Wifi settings received for Viam-2G
2025-06-25T20:34:31.244Z        INFO    viam-agent      networking/networkmanager_linux.go:501  Adding/updating settings for network Viam-2G@wlan0
2025-06-25T20:34:33.284Z        INFO    viam-agent      networking/networkmanager_linux.go:258  Stopping provisioning mode.
2025-06-25T20:34:33.284Z        INFO    viam-agent      networking/networkmanager_linux.go:390  Deactivating connection: viam-setup-10951@wlan0
2025-06-25T20:34:33.295Z        INFO    viam-agent      networking/networkmanager_linux.go:397  Successfully deactivated connection: viam-setup-10951@wlan0
2025-06-25T20:34:33.296Z        INFO    viam-agent      networking/networkmanager_linux.go:280  Stopped hotspot provisioning mode.
2025-06-25T20:34:39.845Z        INFO    viam-agent      networking/networkmanager_linux.go:306  Activating connection: Viam-2G@wlan0
2025-06-25T20:34:43.863Z        INFO    viam-agent      networking/networkmanager_linux.go:356  Successfully activated connection: Viam-2G@wlan0
2025-06-25T20:34:43.977Z        INFO    viam-agent      networking/connstate_linux.go:39        Online: true
2025-06-25T20:34:43.983Z        INFO    viam-agent      networking/connstate_linux.go:65        Wifi Connected: true
```

### DIFF
--- a/subsystems/networking/networkmanager_linux.go
+++ b/subsystems/networking/networkmanager_linux.go
@@ -702,6 +702,7 @@ func (n *Networking) backgroundLoop(ctx context.Context, scanChan chan<- bool) {
 	}
 }
 
+// Process user input (viam.json and/or Wifi settings) and return true if everything succeeded without error, false otherwise.
 func (n *Networking) processUserInput(userInput userInput) bool {
 	if userInput.RawConfig != "" || userInput.PartID != "" {
 		n.logger.Info("Device config received")
--- a/subsystems/networking/networkstate_linux.go
+++ b/subsystems/networking/networkstate_linux.go
@@ -108,7 +108,7 @@ func (n *networkState) genNetKey(ifName, ssid string) string {
 }
 
 // LockingNetwork returns a pointer to a network, wrapped in a lockable struct, so updates are persisted
-// Users must lock the returned network before updates.
+// Users must lock the returned network before updates. Use Network() instead for a read-only copy.
 func (n *networkState) LockingNetwork(iface, ssid string) *lockingNetwork {
 	n.mu.Lock()
 	defer n.mu.Unlock()


subsystems/networking/networkmanager_linux.go
```diff
@@ -750,15 +756,11 @@ func (n *Networking) mainLoop(ctx context.Context) {
 					continue
 				}
 			}
-		default:
```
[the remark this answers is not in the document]
Otterverse: It looks like this would also remove the context cancellation check (the ctx.Done() channel) and I'm unclear why. It's important that any point where we're potentially waiting, a shutdown (via context cancellation) will be obeyed so things don't just hang.

subsystems/networking/networkmanager_linux.go
```diff
@@ -823,12 +825,18 @@ func (n *Networking) mainLoop(ctx context.Context) {
...
 
 			n.logger.Debugf("inactive portal: %t, have candidates: %t, fallback timeout: %t (%s remaining)",
 				inactivePortal, haveCandidates, fallbackHit, fallbackRemaining)
 
-			if shouldExit {
+			if shouldExitPMode {
+				if userInputReceived {
+					// We could get to this point before the user receives our response (poor UX, but likely not critical)
+					// E.g. try to avoid "Not connected" web portal screen.
+					time.Sleep(3 * time.Second)
```
Otterverse, line 838: You should never use regular sleep in agent (or in general in most of Viam's code.) Here, there's a "Sleep()" method on the two healthcheck objects for background and foreground loops that both keeps the healthcheck satisfied and obeys context cancellation. This is the main loop, so probably `n.mainLoopHealth.Sleep(ctx, 3 * time.Second)`

That said, I don't understand why this is needed at all, when it wasn't needed before. I'm pretty sure that the return to any GRPC calls should've already happened before userInputReceived would've been set true.

subsystems/networking/networkmanager_linux.go
```diff
@@ -708,6 +712,7 @@ func (n *Networking) mainLoop(ctx context.Context) {
 
 	scanChan := make(chan bool, 16)
 	inputChan := make(chan userInput, 10)
+	var forceStayInPModeUntilRecvUserInput bool
```
Otterverse, line 715: I'm not sure I understand the purpose of this variable. It already stays in provisioning mode until it receives input OR the long timeout occurs (10 minutes.) The "ShouldExit" calculation below already includes "userInputReceived" as a condition for exiting. Was this some sort of workaround because you reset the lastTried timestamp on the failed network too early above?

subsystems/networking/networkmanager_linux.go
```diff
@@ -861,21 +872,29 @@ func (n *Networking) mainLoop(ctx context.Context) {
 				if n.Config().TurnOnHotspotIfWifiHasNoInternet.Get() {
 					lastConnectivity = n.connState.getLastOnline()
 				}
+			} else {
+				// Did not get what we need in previous provision (e.g. bad wifi password).
+				forceStartPMode = true
```
Otterverse, line 877: How does this distinguish between the regular (old) case of "tryCandidates failed to connect" and "tryCandidates tried a **newly-configured** network (and failed)?"

It seems like it would just always hit this and set force-start to true, whether or not a user had just submitted a new wifi config.

subsystems/networking/networkmanager_linux.go
```diff
@@ -861,21 +872,29 @@ func (n *Networking) mainLoop(ctx context.Context) {
...
 			return
 		}
 
 		hitOfflineTimeout := now.After(lastConnectivity.Add(time.Duration(n.Config().OfflineBeforeStartingHotspotMinutes))) &&
 			now.After(pModeChange.Add(time.Duration(n.Config().OfflineBeforeStartingHotspotMinutes)))
 		// not in provisioning mode, so start it if not configured (/etc/viam.json)
 		// OR as long as we've been offline AND out of provisioning mode for at least OfflineTimeout (2 minute default)
-		if !isConfigured || hitOfflineTimeout {
+		if !isConfigured || hitOfflineTimeout || forceStartPMode {
+			forceStartPMode = false
```
Otterverse, line 897:
```suggestion
		if !isConfigured || hitOfflineTimeout || forceStartPMode {
```
Tiny nit, but the boolean is declared inside the loop, so it already reset to false each iteration when it's redeclared.

Otterverse (changes requested):
Most comments inline. Generally I think this is failing to distinguish between the "bad password" case, and the general case when it's forcing an immediate return to provisioning mode, and that most of these changes are needed. Feel free to slack me tomorrow if you want to discuss or have a quick call!

subsystems/networking/networkmanager_linux.go
Otterverse: What I'm saying is this triggers force start on EVERY failure, not just on newly added networks. That means it just always ignore the timer.

Otterverse (changes requested):
So did some more testing and tried a quick fix of my own. Pretty certain this entire PR can be a one line change for the main issue (adding the sleep for the portal reload, and rearranging the select/default stuff being considered seperately).

```
--- a/subsystems/networking/networkmanager_linux.go
+++ b/subsystems/networking/networkmanager_linux.go
@@ -875,7 +875,7 @@ func (n *Networking) mainLoop(ctx context.Context) {
                        now.After(pModeChange.Add(time.Duration(n.Config().OfflineBeforeStartingHotspotMinutes)))
                // not in provisioning mode, so start it if not configured (/etc/viam.json)
                // OR as long as we've been offline AND out of provisioning mode for at least OfflineTimeout (2 minute default)
-               if !isConfigured || hitOfflineTimeout {
+               if !isConfigured || hitOfflineTimeout || userInputReceived {
                        if err := n.StartProvisioning(ctx, inputChan); err != nil {
                                n.logger.Warn(err)
                        }
```

I tested JUST that one edit on the main branch, and it everything behaves exactly as I would expect/desire. Bad password sends it right back into provisioning mode (because we're on the same loop where we added it.) Otherwise, normal timeouts are obeyed.

The only issue, as has been discussed, is that IF hardware cannot scan while in AP (access point) mode for provisioning, the candidate list will be empty until one or two scan cycles have run, so it'll almost always fail and return to provisioning mode in those cases (though after the next retry timer cycle, it'd eventually work if nothing else is done.)

All that said, I think that risk/reward trade is fine for a quick fix. I also think adding the sleep and rearranging.

As I already rewrote some of it to test and validate the issues I was asserting, I put in a PR against this fork so you don't have to redo work I just did. Feel free to modify if you feel there's more to clean up. PR here: https://github.com/aldenh-viam/agent/pull/1

subsystems/networking/networkmanager_linux.go
```diff
@@ -879,6 +885,25 @@ func (n *Networking) mainLoop(ctx context.Context) {
...
+
+				// In normal mode, tryCandidates will only try a single AP, the PrimarySSID (set in the latest processUserInput).
+				// In roaming mode, all configured connections known to NetworkManager will be tried.
+				// In either case, if we're still offline and the most recent user-provided credentials are bad,
+				// assume it's a mistake and go back to pMode. The lastError will be visible to the user.
+				if nw != nil && nw.lastError != nil && errors.Is(nw.lastError, ErrBadPassword) {
+					n.logger.Warnw("Received incorrect password for AP. Reactivating provisioning mode.",
+						"ssid", userInputSSID, "err", nw.lastError)
+					forceStartPMode = true
+				}
```
Otterverse, line 906: I'm not sure I understand the need for this block (and the forceStartPMode variable below as well.) What's the scenario where this would re-enter provisioning mode but the prior code (from last week, when I approved it) wouldn't? In the comment you said `now only bypasses the timer if user just inputted a wrong password, instead of on any connection failure after user input.` But... we WANT to reenter provisioning in either case, don't we? Lack of connection is still a situation where provisioning is incomplete, so we'll re-enter provisioning anyway. Why wait for the timer in that case if we know there's an active user (who just gave us input) waiting for this?

I'm not trying to reject this outright, but I want to understand the scenario where this makes a difference and understand why that scenario specifically matters, as I feel like maybe I'm not seeing something you are.

subsystems/networking/networkmanager_linux.go
```diff
@@ -879,6 +885,25 @@ func (n *Networking) mainLoop(ctx context.Context) {
...
 				}
+			} else if !connectOk && userInputReceived && userInputSSID != "" {
+				// Special case: received AP credentials and exited provisioning to try it, but it's unsuccessful.
+				// Most commonly a wrong password. In roaming mode, we can get here too if we're connected but offline.
+				// Note: the network tests of `allGood` are updated in the background loop, so we may not get here
+				// in the same iteration as when user input was actually received, but rather a future iteration -
+				// `userInputReceived` and `userInputSSID` will have already been wiped.
+
+				// LockingNetwork is called with these same args in portalSave and tryCandidates; so should exist.
+				nw := n.netState.LockingNetwork(n.Config().HotspotInterface, userInputSSID)
```
Otterverse, line 896: You acquire a locking network, but never actually lock it before reading the fields in it. As noted on the function itself, you should lock the mutex of the returned network object.

Alternately, if you want read-only values (which seems to be the case here), just use .Network() instead of .LockingNetwork()

Otterverse (changes requested):

Otterverse (approved):
As noted in previous comment, there's some stuff to possibly adjust around the timeouts on userinput, but... that's gonna need to work with bluetooth tethering as well, so let's take this PR as-is now, and we/I can tweak further while we test tethering.

subsystems/networking/networkmanager_linux.go
```diff
@@ -702,6 +702,41 @@ func (n *Networking) backgroundLoop(ctx context.Context, scanChan chan<- bool) {
 	}
 }
 
+func (n *Networking) processUserInput(userInput userInput) bool {
```
cheukt, line 705: what does the bool return value represent? the function name is not very descriptive

cheukt (approved):
looks good! just one naming nit

PR #128, opened 2025-07-02
RSDK-11195: systemd KillMode=mixed

aldenh-viam:
Use `KillMode=mixed` instead of the default `control-group` in the systemd service configuration.

`control-group` will send `SIGTERM` to Agent, viam-server, and its child processes (modules).

`mixed` will only send `SIGTERM` to Agent. Agent already takes care of shutting down viam-server, which in turn takes care of shutting down its child processes. If processes are still alive after Agent exits or `TimeoutSec=240`, it will kill them with `FinalKillSignal=SIGQUIT` (both defined in the current configuration).

We have seen some SIGBUS errors emitted from modules when everything is simultaneously `SIGTERM`'ed instead of following the graceful shutdown sequence. This change will help avoid that.

This change does not automatically update an installed `viam-agent.service`. Do we need to advise users to re-run `viam-agent --install`?

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,3 +19,8 @@ Ex: `make all TAG_VERSION=0.1.2`
 
 ### DevMode
 Setting the environment variable `VIAM_AGENT_DEVMODE=1` will skip the self-location check for the binary, so you can run it directly during development, without installing to /opt/viam.
+
+### Systemd
+In production, Agent is designed to run as a systemd service. First-time setup is `sudo ./viam-agent --install` and it will automatically update itself when new updates are released.
+
+The service configration lives in both `viam-agent.service` and `preinstall.sh`, and the two should be kept in sync when making changes.


cheukt:
isn't InstallNewVersion run for all new versions downloaded? https://github.com/viamrobotics/agent/blob/dcb688946943af915fbafce63cb2c64ea612e8e5/agent.go#L33

Otterverse:
> isn't InstallNewVersion run for all new versions downloaded?
> 
> https://github.com/viamrobotics/agent/blob/dcb688946943af915fbafce63cb2c64ea612e8e5/agent.go#L33
Confirming this should get installed during update automatically.

That said, there is a "hidden" copy of the service file embedded in preinstall.sh that should be updated to match.

viam-agent.service
```diff
@@ -11,6 +11,7 @@ RestartSec=5
 User=root
 TimeoutSec=240
 ExecStart=/opt/viam/bin/viam-agent --config /etc/viam.json
+KillMode=mixed
```
ale7714, line 14: Please, add a comment that this file must be kept in sync with the preinstall script. Thank you

PR #132, opened 2025-07-18
[APP-8900] Make downloads avoid accidental overwrites, add tests

Otterverse:
The cache (json) format used by Agent changed a few major versions ago. Jumping from an old version to the current version CAN result in the running agent binary not being in the cache file, and thus when a download for it is attempted, it can over overwrite the running file. If this is interrupted before the process is complete (e.g. the machine is restarted or there's a network outage for a while) it can leave the agent binary in a broken state, and it cannot start back up.

This PR fixes that by returning early from DownloadFile() if the actual io.Copy() fails (previously this wasn't caught until AFTER the temp file was renamed to replace the binary.) Additional code is also added to modify the output path name if a duplicate is detected. Lastly, tests are added for this.

As part of testing, this uncovered a race condition in the progress bar library, so a workaround with our own locking is implemented as well.

### DIFF
--- a/utils/utils_test.go
+++ b/utils/utils_test.go
@@ -1,10 +1,18 @@
 package utils
 
 import (
+	"context"
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"os"
 	"os/exec"
 	"path/filepath"
+	"strings"
 	"testing"
+	"time"
 
+	"go.viam.com/rdk/logging"
 	"go.viam.com/test"
 )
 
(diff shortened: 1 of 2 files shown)


utils/utils_test.go
```diff
@@ -94,3 +102,198 @@ func TestWriteFileIfNew(t *testing.T) {
 	test.That(t, err, test.ShouldBeNil)
 	test.That(t, written, test.ShouldBeTrue)
 }
+
+func TestDownloadFile(t *testing.T) {
```
Otterverse, line 106: I tried a new thing, and let AI write the bulk of this test, and tweaked a few things manually afterward. It SEEMS to be good, but I'm still hesitant about AI.

Otterverse:
> Small non-blockers:
> 
> 1. Would be good to clean up the 100 invalid downloads sometime, if it's not done already. ~23MB * 100 is not nothing.

It should really never get to that many (nor hopefully ever use this at all!) But... cleanup code already exists. There's a "CleanCache()" function in version_control.go which should clean up older versions AND untracked (aka duplicate) versions/files eventually, and it has safety checks to make sure the files it removes aren't currently symlinked/in-use.

> 2. Are we sure `io.Copy` is what errored? I haven't seen the machine logs. Also the `if errRet != nil {...}` early return won't include errors from the deferred blocks calling `.Close()`, in case those might be useful to act on as well.

I did not have a way to perfectly reproduce the reported error, so I can't say with 100% certainty, but... io.Copy is where the actual "download" takes place (or at least the "write to disk part"), and it definitely did not have an early return before the next step, where it moves/renames the temp file to replace the previous real file. That leaves me pretty confident it MUST be what errored, but if you have any other theories or ideas (or tests to implement to prove/disprove it) please let me know.

As for the deferred blocks not including errors... I'm not sure I see where you're talking about. errRet is pre-declared as an output/return value, thus all the defer blocks should have access to it when they fire, and as such, they're all modifying it by joining new errors to it, meaning ALL errors should get returned (as a joined multi-error.) If I wrong about that or overlooked something, please do point it out.

> 3. Curious: what's the race condition in the progress bar library?

Basically you can call bar.String() (to get a rendered version of the progress bar from the internal ".rendered" field) WHILE another go routine might actually be rendering an update and writing it there. No locks on that data structure in the progressbar itself.

PR #133, opened 2025-07-29
Fix typo

npentrel:


Otterverse:
About to cut a release, so pushing this in before that!

PR #138, opened 2025-09-04
Avoid returning early from checkConnections() loop.

Otterverse:
If other/virtual NetworkManager devices "go away" (which can happen with DBUS) the loop could exit early and not find the device(s) we care about. This should avoid that, and merely log a warning instead, and remove the "dead" device.

### DIFF
--- a/subsystems/networking/networkmanager_linux.go
+++ b/subsystems/networking/networkmanager_linux.go
@@ -12,6 +12,7 @@ import (
 	"reflect"
 	"slices"
 	"sort"
+	"strings"
 	"time"
 
 	gnm "github.com/Otterverse/gonetworkmanager/v2"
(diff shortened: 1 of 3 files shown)


subsystems/networking/networkmanager_linux.go
```diff
@@ -128,7 +129,17 @@ func (n *Networking) checkOnline(ctx context.Context, force bool) error {
...
 
-func (n *Networking) checkConnections() error {
+// isObjectNotExistError checks if the error indicates that an object no longer exists at the given path.
+func isObjectNotExistError(err error) bool {
+	if err == nil {
+		return false
+	}
+
+	// Fallback to string matching as dbus doesn't return "true" Go-like enumerated errors, and has it's own string-based error type.
+	return strings.Contains(err.Error(), "Object does not exist at path")
```
abe-winter, line 139: this isn't super helpful (I don't see how we can reference it directly), but I think this error comes from here:
https://github.com/GNOME/glib/blob/8a6f6b352e95a0c93f9af5789e108c50ad6d785c/gio/gdbusconnection.c#L7879

or elsewhere in that codebase

abe-winter (approved):

PR #139, opened 2025-09-10
APP-8893 Fix redundant 'viam-server reports it is unsafe to restart' log

benjirewis:
[WIP] Still want to test on a pi once I get into my coworking space today.

### DIFF
--- a/subsystems/viamserver/viamserver.go
+++ b/subsystems/viamserver/viamserver.go
@@ -247,6 +247,11 @@ func (s *viamServer) HealthCheck(ctx context.Context) error {
 	return nil
 }
 
+// errUnsafeToRestart is reported to a result channel in isRestartAllowed below when any
+// one of the viam-server test URLs explicitly report that it is unsafe to start
+// the viam-server instance.
+var errUnsafeToRestart = errors.New("viam-server reports it is unsafe to restart")
+
 // Must be called with `s.mu` held, as `s.checkURL` and `s.checkURLAlt` are
 // both accessed.
 func (s *viamServer) isRestartAllowed(ctx context.Context) (bool, error) {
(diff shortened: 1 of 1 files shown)


subsystems/viamserver/viamserver.go
```diff
@@ -304,15 +309,25 @@ func (s *viamServer) isRestartAllowed(ctx context.Context) (bool, error) {
...
+			resultChan <- errUnsafeToRestart
 		}()
 	}
+
 	var combinedErr error
 	for i := 1; i <= len(urls); i++ {
 		result := <-resultChan
+
+		// If any test URL reports it is explicitly _safe_ to restart (nil value sent to
+		// resultChan), we can assume we can restart viam-server.
```
benjirewis, line 321: Correct me if I'm wrong, but this comment is meant to describe the current behavior of testing against all "test URLs" here.

subsystems/viamserver/viamserver.go
```diff
@@ -304,15 +309,25 @@ func (s *viamServer) isRestartAllowed(ctx context.Context) (bool, error) {
...
 		if result == nil {
 			return true, nil
 		}
+		// If any test URL reports it is explicitly _unsafe_ to restart (errUnsafeToRestart),
+		// we can assume we should not restart viam-server.
+		if errors.Is(err, errUnsafeToRestart) {
```
benjirewis, line 327: I'm fairly certain a log like:
```
7/18/2025, 4:07:20 PM warn viam-agent   viamserver/viamserver.go:349   viam-server reports it is unsafe to restart viam-server reports it is unsafe to restart viam-server reports it is unsafe to restart viam-server reports it is unsafe to restart viam-server reports it is unsafe to restart
```
Is caused by five test URLs pointing to the same `viam-server` instance _all_ reporting that `viam-server` is unsafe to restart. I do not think we want to combine the errors in that case, and can, conservatively, return `false, errUnsafeToRestart` in the event that any one of the test URLs reports `RestartAllowed: false`.

PR #149, opened 2025-09-22
RSDK-11835: Prevent redundant self update after initial install

jmatth:
This fixes an issue where the agent will always perform a self update after initial install by re-downloading it's binary and restarting. This happened because the install process run by `viam-agent --install` just symlinks itself to `/opt/viam/bin/viam-agent` and sets up the systemd unit. In order to  recognize an install as valid the agent expects to see certain values in the version cache, which was not created, and the symlink for the agent pointing to a specific directory, which was not true. This resolves these issues by updating the install process to copy the binary into the expected location at `/opt/viam/cache`, symlink to that copy, and write out a minimal version cache to disk. The rest of the version information is merged in when it is fetched from app on first agent run. Testing was done manually with a binary that was slightly modified to ignore the hash mismatch.

### DIFF
--- a/utils/utils.go
+++ b/utils/utils.go
@@ -574,3 +574,30 @@ func downloadProgressSetup(ctx context.Context,
 	}()
 	return writer, cancel
 }
+
+// AtomicCopy implements a best effort to atomically copy the file at src to
+// dst. It does this by copying first to a temporary file in the same directory
+// as dst, then renaming that file to the final expected path.
+func AtomicCopy(dst, src string) error {
+	infile, err := os.Open(src)
+	if err != nil {
+		return errw.Wrap(err, "opening source file for atomic copy")
+	}
+	defer infile.Close()
+	tmpDst := dst + ".tmp"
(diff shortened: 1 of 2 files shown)


utils/utils.go
```diff
@@ -574,3 +574,48 @@ func downloadProgressSetup(ctx context.Context,
 	}()
 	return writer, cancel
 }
+
+// AtomicCopy implements a best effort to atomically copy the file at src to
+// dst. It does this by copying first to a temporary file in the same directory
+// as dst, then renaming that file to the final expected path.
```
benjirewis, line 580: [q] Why is an atomic copy necessary? Do you expect there to be some other, concurrent creator of an identically-named agent binary in cache? Or, is this just a best principles kind of thing?

jmatth: Just a best practices thing. This is how a lot of system management programs like Chef handle managed files because it prevents leaving the file in a partially copied state if the program or OS crashes mid copy and works across file systems.

PR #153, opened 2025-09-26
RSDK-11248 Add restart checking logic

benjirewis:
[RSDK-11248](https://viam.atlassian.net/browse/RSDK-11248)

Adds similar restart checking logic to the logic removed in https://github.com/viamrobotics/rdk/pull/5324; the logic now restarts `viam-agent` and all its other subsystems along with `viam-server`. Refactors hits of `restart_status` to be more generic (can query multiple restart "properties"). Adds a `Property` method to the `Subsystem` interface.

## Testing

I've done a decent amount of manual testing but still want to do more to ensure I haven't broken the "restart allowed" logic nor the behavior of restarts on Windows.

Note that I'm leveraging a `does_not_handle_needs_restart` JSON property exposed on the `restart_status` endpoint introduced in the associated RDK PR. The expected behavior right now is:

|                 | Old viam-agent                             | New viam-agent                             |
|-----------------|--------------------------------------------|--------------------------------------------|
| Old viam-server | viam-server should handle restart checking | viam-server should handle restart checking |
| New viam-server | neither will handle restart checking       | viam-agent should handle restart checking  |

### DIFF
--- a/subsystems/viamserver/viamserver.go
+++ b/subsystems/viamserver/viamserver.go
@@ -44,6 +44,12 @@ type viamServer struct {
 	// for blocking start/stop/check ops while another is in progress
 	startStopMu sync.Mutex
 
+	// whether this viamserver instance handles needs restart checking itself; calculated
+	// and cached at startup; used by the manager to determine whether agent should handle
+	// needs restart checking on viamserver's behalf (this is the case for new viamserver
+	// versions)
+	doesNotHandleNeedsRestart bool
+
 	logger logging.Logger
 }
 
@@ -156,6 +162,21 @@ func (s *viamServer) Start(ctx context.Context) error {
 		s.checkURLAlt = strings.Replace(matches[2], "0.0.0.0", "localhost", 1)
 		s.logger.Infof("viam-server restart allowed check URLs: %s %s", s.checkURL, s.checkURLAlt)
 		s.logger.Infof("%s started", SubsysName)
+
+		// Once the subsystem has successfully started, check whether it handles needs restart
+		// logic. We can calculate this value only once at startup and cache it, with the
+		// assumption that it will not change over the course of the lifetime of the
+		// subsystem.
+		s.mu.Lock()
+		s.doesNotHandleNeedsRestart, err = s.checkRestartProperty(ctx, RestartPropertyDoesNotHandleNeedsRestart)
+		s.mu.Unlock()
+		if err != nil {
+			s.logger.Warn(err)
+		}
+		if !s.doesNotHandleNeedsRestart {
+			s.logger.Warnf("%s may already handle checking needs restart functionality; will not handle in agent",
+				SubsysName)
+		}
 		return nil
 	case <-ctx.Done():
 		return ctx.Err()
@@ -245,6 +266,34 @@ func (s *viamServer) Update(ctx context.Context, cfg utils.AgentConfig) (needRes
 		return true
 	}
 	return false
+}
+
+// Property returns a single property of the currently running viamserver.
+func (s *viamServer) Property(ctx context.Context, property string) bool {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	switch property {
+	case RestartPropertyRestartAllowed:
+		if !s.running || runtime.GOOS == "windows" {
+			// Assume agent can restart viamserver if the subsystem is not running or we are
+			// running on Windows.
+			return true
+		}
+
+		restartAllowed, err := s.checkRestartProperty(ctx, RestartPropertyRestartAllowed)
+		if err != nil {
+			s.logger.Warn(err)
+		}
+		return restartAllowed
+	case RestartPropertyDoesNotHandleNeedsRestart:
+		// We can use the cached value (calculated in Start) for handle needs restart
+		// property.
+		return s.doesNotHandleNeedsRestart
+	default:
+		s.logger.Errorw("Unknown property requested from viamserver", "property", property)
+		return false
+	}
 }
 
 func NewSubsystem(ctx context.Context, logger logging.Logger, cfg utils.AgentConfig) subsystems.Subsystem {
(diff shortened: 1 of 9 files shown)


subsystems/viamserver/viamserver.go
```diff
@@ -59,9 +45,22 @@ type viamServer struct {
...
+
 	logger logging.Logger
 }
 
+type ctxKey int
+
+// CtxKeySendSIGQUITInsteadOfSIGTERM is used by manager to signal to the Stop method that
+// SIGQUIT should be sent instead of SIGTERM to get a goroutine dump and stop immediately
+// (when app has forcibly restarted agent).
+const CtxKeySendSIGQUITInsteadOfSIGTERM = ctxKey(iota)
```
benjirewis, line 62: I don't _love_ this solution to tell `viam-server` to `Stop` using SIGQUIT instead of SIGTERM, but I didn't want to further change the `Subsystem` interface.

version_control.go
```diff
@@ -216,8 +216,8 @@ func (c *VersionCache) UpdateBinary(ctx context.Context, binary string) (bool, e
 	shasum, err := utils.GetFileSum(verData.UnpackedPath)
 	if err == nil {
 		goodBytes = bytes.Equal(shasum, verData.UnpackedSHA)
-	} else {
-		c.logger.Warn(err)
+	} else if verData.UnpackedPath != "" { // custom file:// URLs with have an empty unpacked path; no need to warn
```
benjirewis, line 219: [drive-by] Logging fix related to comment.

subsystems/viamserver/viamserver.go
```diff
@@ -150,16 +149,19 @@ func (s *viamServer) Start(ctx context.Context) error {
...
-		}
-		if s.cmd.ProcessState != nil {
-			s.lastExit = s.cmd.ProcessState.ExitCode()
-			if s.lastExit != 0 {
-				s.logger.Errorf("non-zero exit code: %d", s.lastExit)
-			}
-		}
+		// Only log errors from Wait() or the exit code of the process state if subsystem
+		// exited unexpectly (was not stopped by agent and is therfore still marked as
+		// shouldRun).
```
benjirewis, line 154: We were getting unnecessary error logs here when restarting. A non-zero exit code when we sent SIGQUIT to `viam-server` is expected.

cmd/viam-agent/main.go
```diff
@@ -211,7 +211,8 @@ func setupExitSignalHandling() (context.Context, context.CancelFunc) {
 			case syscall.SIGABRT:
 				fallthrough
 			case syscall.SIGTERM:
-				globalLogger.Info("exiting")
+				globalLogger.Infof("Signal %s was received. %s will now exit to be restarted by service manager",
```
benjirewis, line 214: This change, and the addition and usage of `reason` for the `Exit` method in `manager.go` are for https://viam.atlassian.net/browse/RSDK-11266. Hopefully these messages will be useful for debugging.

subsystems/viamserver/viamserver.go
```diff
@@ -168,9 +181,24 @@ func (s *viamServer) Start(ctx context.Context) error {
 	select {
 	case matches := <-c:
 		s.checkURL = matches[1]
-		s.checkURLAlt = strings.Replace(matches[2], "0.0.0.0", "localhost", 1)
-		s.logger.Infof("viam-server restart allowed check URLs: %s %s", s.checkURL, s.checkURLAlt)
+		s.checkURLAlt = strings.Replace(matches[2], "0.0.0.0", "127.0.0.1", 1)
```
benjirewis, line 184: I changed this to `127.0.0.1` so checks to `checkURLAlt` would work on Windows (`localhost` does not exist on Windows) as well as Linux. Restart allowed checks _could_ already work with the `checkURL` (this was something like `https://vader-main.ydbecmp2sz.local.viam.cloud:8080`).

subsystems/viamserver/viamserver.go
```diff
@@ -348,28 +300,31 @@ func (s *viamServer) Update(ctx context.Context, cfg utils.AgentConfig) (needRes
...
 	s.mu.Lock()
 	defer s.mu.Unlock()
 
-	if !s.running || runtime.GOOS == "windows" {
-		return true
-	}
+	switch property {
+	case RestartPropertyRestartAllowed:
+		if !s.running {
+			// Assume agent can restart viamserver if the subsystem is not running.
```
benjirewis, line 311: We used to also `return true` unconditionally on Windows, meaning a maintenance sensor returning `false` or an active reconfiguration would not stop `viam-agent` from restarting `viam-server` for an update on Windows. That's no longer the case, and both of those things will prohibit `viam-agent` with these changes from restarting `viam-server` (the `WARN` log `viam-server has NOT allowed a restart; will NOT restart` will be output).

utils/utils_windows.go
```diff
@@ -55,6 +55,14 @@ func SignalForTermination(pid int) error {
 	return windows.GenerateConsoleCtrlEvent(syscall.CTRL_BREAK_EVENT, uint32(pid)) //nolint:gosec
 }
 
+func SignalForQuit(pid int) error {
+	// TODO(Benji): Grab a goroutine dump from "http://localhost:8080/debug/pprof/heap"
+	// but that only works when EnableWebProfile == true.
```
benjirewis, line 60: Per the TODO: with my changes here, restarting a part _on Windows_ no longer outputs a goroutine dump from `viam-server`. That was a helpful feature for when `viam-server` was borked and we wanted to both restart it and see what goroutines got stuck. I'm using SIGQUIT to do this on Linux, but SIGQUIT on Windows doesn't exist. If we still want stack traces, we need another solution here, but I left it as a TODO because I wasn't sure how critical it was.

subsystems/viamserver/viamserver.go
benjirewis: I've put the hard-coding back for now and filed https://viam.atlassian.net/browse/RSDK-12271 to re-enable at some point in the future. Re-enabling a feature that caused a lot of Windows issues previously as part of this already-large PR is probably too large of a risk.

PR #154, opened 2025-09-29
Prevent redundant self update after initial install mk2

jmatth:


agent.go
```diff
@@ -50,80 +55,76 @@ func InstallNewVersion(ctx context.Context, logger logging.Logger) (bool, error)
...
+			return errw.Wrapf(err, "installing symlink at %s", expectedBinPath)
 		}
-	}
 
-	if newFile {
-		cmd = exec.Command("systemctl", "daemon-reload")
-		output, err = cmd.CombinedOutput()
-		if err != nil {
-			return errw.Wrapf(err, "running 'systemctl daemon-reload' output: %s", output)
+		if !VersionCacheExists() {
```
jmatth, line 101: This conditional is the only functional change from the last PR. In the case of a self update the previous version of the agent that's performing the update handles updating the cache, so we don't need to touch it.

cheukt (approved):
LGTM, thanks for adding tests! maybe kind of a dumb q, but I don't think I see a new test that tests the issue that caused the revert? did I miss it?

jmatth:
Oops, thanks for catching that. Tests for fresh installs + self updates are in [`agent_test.go`][1], which I forgot to commit 😞

[1]: https://github.com/jmatth/viam-agent/blob/0f17ecbac7580378d43e65280ba2d940da8b594a/agent_test.go

utils/systemd/manager.go
```diff
@@ -0,0 +1,169 @@
...
+		s.logger.Warn("If you customized this file, please run 'systemctl edit viam-agent' and create overrides there")
+		if err := os.RemoveAll(oldPath); err != nil {
+			s.logger.Warn(errors.Wrapf(err, "removing old service file %s, please delete manually", oldPath))
+		}
+	}
+
+	if newFile {
+		if err := s.DaemonReload(); err != nil {
+			return "", false, err
+		}
```
benjirewis, line 113: Are you missing `s.Enable()` here? Trying to follow along [from where](https://github.com/viamrobotics/agent/blob/8a6c789ff1633262cfb68baf3249f6e2f1e1fb69/agent.go#L121-L129) you moved this code.

utils/systemd/manager_test.go
```diff
@@ -0,0 +1,164 @@
...
+			if tc.includeNewSearchPath {
+				test.That(t, serviceFile, test.ShouldEqual, filepath.Join(defaultServiceDir, "my-service.service"))
+			} else {
+				test.That(t, serviceFile, test.ShouldEqual, filepath.Join(fallbackServiceDir, "my-service.service"))
+			}
+		})
+	}
+
+	t.Run("new install", func(t *testing.T) {
+	})
```
benjirewis, line 137: ?

utils/systemd/manager.go
```diff
@@ -0,0 +1,169 @@
...
+}
+
+// Annoying workaround to allow embedding SystemdExecutor in SystemdManager w/o
+// allowing it to be modified from outside the module.
+type privateExecutor = SystemdExecutor
+
+// SystemdManager provides methods for making high-level changes to systemd
+// services.
+type SystemdManager struct {
+	privateExecutor
```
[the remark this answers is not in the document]
jmatth: If I did that I would have to write an annoying wrapper function to expose any methods from `SystemdExecutor` that need to be accessed by users of `SystemdManager`:

```go
func (s *SystemdManager) IsAvailable() error {
	return s.systemdExecutor.IsAvailable()
}

// Repeat above for all relevant methods.
```

Alternatively we could just embed the entire executor interface in the manager struct:

```go
type SystemdManager struct {
	SystemdExecutor
	dirs   systemdDirs
	logger logging.Logger
}
```

...but makes the field exported and suddenly other models can start modifying what should be an internal field:

```go
// Contrived example of clearly bad code but why allow it if we don't need to?
sdmanager.SystemdExecutor = nil
```

This gives us the best of both worlds: the interface is embedded so its methods are automatically included on `SystemdManager` but its name is lowercase so it can't be messed with outside the module.

version_control.go
```diff
@@ -26,9 +26,14 @@ import (
...
-const (
-	versionCacheFilename = "version_cache.json"
-)
+func getCacheFilePath() string {
+	return filepath.Join(utils.ViamDirs.Cache, "version_cache.json")
+}
+
+func VersionCacheExists() bool {
+	_, err := os.Stat(getCacheFilePath())
+	return !errors.Is(err, os.ErrNotExist)
```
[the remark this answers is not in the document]
jmatth: Good point, I think I cargo culted this on from somewhere else.

PR #215, opened 2026-03-18
APP-15718 MSI installer and non-SYSTEM account support on windows

abe-winter:
## What changed
- add `Package.wxs` which builds an msi for installing agent on windows
- installer UI allows users to choose root path and non-SYSTEM account
- config-path is also configurable but only from msiexec command line for now
- VIAM_HOME is set for the viam-agent service when root path is overridden
## Why
Support more windows targets
## Testing
We're able to install and start the win-reg-sensor module on win11 with:
- default msi settings
- custom root on C:
- custom root on D:
- uninstall-reinstall: misbehaves unless you delete the viam root folder first. I think this a bug in the download logic, I'll ticket. otherwise works.
- non-SYSTEM; also checked that RDK stable -> dev worked

Also tested normal and non-SYSTEM installs on win10. The uninstaller may hang on win10; we should test, but because uninstall is a new feature, this isn't a merge blocker.

This is the config I used:

```json
{
  "components": [
    {
      "name": "sensor-1",
      "api": "rdk:component:sensor",
      "model": "viam:win-reg-sensor:registry",
      "attributes": {
        "keys": [
          "SOFTWARE\\Microsoft\\Windows NT\\CurrentVersion:ProductName"
        ]
      }
    }
  ],
  "modules": [
    {
       "type": "registry",
       "name": "win-reg-sensor",
       "module_id": "viam:win-reg-sensor",
       "version": "0.0.8"
    }
  ]
}
```
## Follow-up work
Remove the ps1 / exe installer, build + upload the msi in CI, update docs to refer to the msi.


windows-installer-agent.ps1
```diff
@@ -1,7 +1,11 @@
 # Viam Agent Installation Script for Windows
 [CmdletBinding()]
 param(
-    [switch]$Silent = $false
+    [switch]$Silent = $false,
+    [string]$RootPath = "C:\opt\viam",
```
ale7714, line 5: ~i thought we want to not use `C:`~ nvm i get it now

windows-installer-agent.ps1
```diff
@@ -109,40 +131,150 @@ try {
...
+    New-NetFirewallRule -Name "viam-agent" -DisplayName "Viam Agent" `
+        -Program $agentBinPath -Direction Inbound -Action Allow -Enabled True | Out-Null
 } catch {
     Write-Warning "Failed to configure firewall: $_"
     # Continue despite firewall error
 }
+
+# Add an ACE to a Windows service's DACL for the given account.
+# Used to grant non-SYSTEM accounts specific service permissions.
+function Add-ServiceDaclAce {
```
ale7714: i'm not sure i fully understand what this does

windows-installer-agent.ps1
```diff
@@ -109,40 +131,150 @@ try {
...
+        $cfgAcl = Get-Acl $configDir
+        $cfgRule = New-Object System.Security.AccessControl.FileSystemAccessRule(
+            $svcAccountName, "ReadAndExecute", "ContainerInherit,ObjectInherit", "None", "Allow")
+        $cfgAcl.AddAccessRule($cfgRule)
+        Set-Acl -Path $configDir -AclObject $cfgAcl
+
+        # Grant SeCreateSymbolicLinkPrivilege so the agent can create symlinks
+        # at runtime when downloading new versions of itself and viam-server.
+        # secedit uses *SID format (not account names), so resolve first.
+        if (-not $Silent) { Write-Host "  Granting $svcAccountName symlink creation privilege..." }
```
ale7714, line 200: is `viam-server` able to update modules because it inherits same permissions?

ale7714 (commented):
looks overall right but i'm not a windows expert. i think if you have test it and it works ✅ . my only concern would be wether we broke something for the system account path. did we test that as well?

abe-winter:
> wether we broke something for the system account path

^ I tested but I'll retest the existing installer on our targets before merging

windows-installer-agent.ps1
abe-winter: these perms are on the account, not the binary, so yes the RDK subproc has them

but also I don't think RDK symlinks during module installation; the ACL on the `$RootPath` matters more for RDK

windows-installer-agent.ps1
abe-winter: we call `Add-ServiceDaclAce(viam-agent, NT_SERVICE\viam-agent, LCRPWPLORC)`

on L269 below, this gives the viam-agent service account permission to manage the viam-agent service

... but there's a chance these perms are available automatically

Also I think the L264 call (firewall perms) is not giving us firewall edit permissions, and the `allowFirewall` implementation in utils.go is swallowing the error

I'm moving this back to draft PR and will test a bit

cheukt:
ran an install to an e drive, seems to work quite well!

ale7714 (approved):

ale7714:
@abe-winter are you keeping the ps1 in case the msi doesn't work?

abe-winter:
> keeping the ps1 in case the msi doesn't work

I wanted to update docs / CI before removing the exe, and I also wanted to keep the wix file separate because it's legitimately hard to write + read

PR #222, opened 2026-03-30
APP-15754, RSDK-13740 - macOS agents orphanes viam-server during agent updates

cheukt:
More background on the bug and why the changes fixes them. I believe the summary of the causes are correct, open to ideas if any of the solutions seem unclean!

### Root Cause

On macOS, `InstallService()` was unconditionally calling `launchctl kickstart -k` after every install **and** update. The `-k` flag sends SIGTERM to the running agent's process group before it can coordinate a clean viam-server shutdown, canceling the agent's global context mid-`SubsystemUpdates`.

On Linux, `systemd.InstallService()` only writes unit files — it never kills the running agent. The agent controls its own exit via `m.Exit()` and launchd's `KeepAlive: true` restarts it, which is the correct behavior.

---

### Bug 1: Orphan viam-server causes port conflict on restart

With the global ctx canceled mid-`SubsystemUpdates`:

1. `Stop()` holds `startStopMu` while waiting for viam-server to exit
2. A concurrent `Start()` call blocks on `startStopMu`
3. `Stop()` finishes and releases the lock
4. `Start()` unblocks — spawns viam-server via `s.cmd.Start()` **before** checking `ctx.Done()`, then returns immediately when it sees the canceled ctx
5. Orphan viam-server runs untracked and claims port 8080
6. New agent's viam-server fails to start due to port conflict

**Fix (`subsystems/viamserver/viamserver.go`):** Check `ctx.Err()` immediately after acquiring `startStopMu`. If ctx is already canceled, return before spawning any process.

---

### Bug 2: Restart coordination skipped on canceled ctx

With ctx canceled, the `RestartAllowed` + `Stop` + `Exit` path fails immediately, causing the code to skip the clean `m.Exit()` path and fall through to `CloseAll`, which races with the new agent process starting.

**Fix (`manager.go`):** At the restart coordination block, check `ctx.Err()` first. If canceled, skip `RestartAllowed` — we know the agent is shutting down — stop viam-server unconditionally with a fresh context, then call `m.Exit()`.

---

### Bug 3: `--install` subprocess killed by bootout before it can kickstart the new agent

`InstallNewVersion` used `exec.CommandContext(ctx, ...)` + `CombinedOutput()`, blocking on the subprocess. On macOS, the subprocess calls `Bootout` internally, which kills the agent's process group (including the subprocess) before it can finish bootstrapping and kickstarting the new agent.

**Fix (`agent_darwin.go`):** Darwin-specific implementation uses `exec.Command` (no ctx), `Setpgid: true` so the subprocess runs in its own process group and survives the bootout, and `cmd.Start()` + a goroutine for `Wait()`. A `select` on `ctx.Done()` vs a done channel handles both cases: plist unchanged (subprocess exits cleanly) and plist changed (ctx canceled by bootout — log and return, the subprocess will kickstart the new agent).

---

### Fix 4: Stop kickstarting on updates; kickstart after bootout (`utils/launchd/manager.go`)

Changed kickstart condition from `!installed` to `needsBootstrap`. For unchanged existing installs, no kickstart is needed — `KeepAlive` handles restart automatically. For new installs and plist-changed updates (where `Bootout` removed the service from launchd's registry), kickstart is required to start the new agent. This mirrors Linux behavior where the installer never kills the running agent.

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -8,7 +8,6 @@ import (
 	"fmt"
 	"io/fs"
 	"os"
-	"os/exec"
 	"path/filepath"
 	"runtime"
 	"strings"
@@ -25,26 +24,6 @@ type systemManager interface {
 	IsAvailable(ctx context.Context) error
 	InstallService(ctx context.Context, serviceName string, serviceFileContents []byte) (string, bool, error)
 	Enable(ctx context.Context, serviceName string) error
-}
-
-// InstallNewVersion runs the newly downloaded binary's Install() for installation of systemd files and the like.
-func InstallNewVersion(ctx context.Context, logger logging.Logger) (bool, error) {
-	if runtime.GOOS == "windows" {
-		// windows doesn't have systemctl so we don't do a postinstall yet.
-		return true, nil
-	}
-	expectedPath := filepath.Join(utils.ViamDirs.Bin, SubsystemName)
-
-	// Run the newly updated version to install systemd and other service files.
-	//nolint:gosec
-	cmd := exec.CommandContext(ctx, expectedPath, "--install")
-	output, err := cmd.CombinedOutput()
-	logger.Info("running viam-agent --install for new version")
-	logger.Info(string(output))
-	if err != nil {
-		return false, errw.Wrapf(err, "running post install step %s", output)
-	}
-	return true, nil
 }
 
 // Install is directly executed from main() when --install is passed.
--- a/agent_darwin.go
+++ b/agent_darwin.go
@@ -1,7 +1,17 @@
 package agent
 
 import (
+	"context"
 	_ "embed"
+	"errors"
+	"os/exec"
+	"path/filepath"
+	"time"
+
+	errw "github.com/pkg/errors"
+	"github.com/viamrobotics/agent/utils"
+	"go.viam.com/rdk/logging"
+	rutils "go.viam.com/rdk/utils"
 )
 
 const (
@@ -10,3 +20,63 @@ const (
 
 //go:embed com.viam.agent.plist
 var serviceFileContents []byte
+
+// InstallNewVersion runs the newly downloaded binary's Install() for installation of launchd service files and the like.
+func InstallNewVersion(ctx context.Context, logger logging.Logger) (bool, error) {
+	expectedPath := filepath.Join(utils.ViamDirs.Bin, SubsystemName)
+
+	// Run the newly updated version to install launchd service files.
+	logger.Info("running viam-agent --install for new version")
+	// On macOS, --install may triggers a launchd bootout which kills the running agent's
+	// process group before the subprocess can finish bootstrapping and kickstarting the
+	// new agent. So start the subprocess in its own process group.
+	//nolint:gosec
+	cleanup := rutils.SlowLogger(
+		ctx,
+		"Waiting for new version of viam-agent to finish installing", "", "",
+		logger,
+	)
+	defer cleanup()
+	ctx, cancel := context.WithTimeout(ctx, 4*time.Minute)
+	defer cancel()
+
+	cmd := exec.Command(expectedPath, "--install")
+	utils.PlatformProcSettings(cmd)
+
+	if err := cmd.Start(); err != nil {
+		return false, errw.Wrap(err, "error running install step")
+	}
+
+	// We intentionally start the Wait() in a goroutine. It is expected that if the plist file
+	// is updated, the installer will trigger a bootout, which will kill this agent process before the
+	// installer process exits. In that case, this goroutine will be leaked but will exit when the installer completes.
+	//
+	// If the plist file is unchanged, this goroutine will wait for the installer to exit and then return cleanly.
+	doneCh := make(chan struct{})
+	go func() {
+		if err := cmd.Wait(); err != nil {
+			logger.Warnw("error while waiting for --install subprocess to complete", "err", err)
+		}
+		close(doneCh)
+	}()
+
+	select {
+	case <-ctx.Done():
+		switch {
+		case errors.Is(ctx.Err(), context.Canceled):
+			// The common case is that the installer will bootstrap and kickstart the new version of agent after this agent process
+			// exits.
+			//
+			// Technically this could also happen if the subprocess had just kicked off and a user/process SIGTERMs the agent, but barring
+			// something very wrong, the agent will still be restarted by launchd and re-attempt the update.
+			logger.Info("viam-agent shutdown requested, will be restarted by the installer or launchd")
+		case errors.Is(ctx.Err(), context.DeadlineExceeded):
+			logger.Warn("viam-agent timed out while updating, will exit and be restarted by launchd")
+		default:
+			logger.Warnw("unexpected context error while updating, viam-agent will exit and be restarted by launchd", "error", ctx.Err())
+		}
+	case <-doneCh:
+		logger.Info("viam-agent update completed, will be restarted by launchd")
+	}
+	return true, nil
+}
--- a/agent_darwin.go
+++ b/agent_darwin.go
@@ -6,10 +6,10 @@ import (
 	"errors"
 	"os/exec"
 	"path/filepath"
-	"time"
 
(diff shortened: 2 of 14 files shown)


jmatth (approved):

cheukt:
@jmatth made a decent amount of changes to the pr, so re-requesting reviews

agent_darwin.go
```diff
@@ -10,3 +20,63 @@ const (
...
+	// installer process exits. In that case, this goroutine will be leaked but will exit when the installer completes.
+	//
+	// If the plist file is unchanged, this goroutine will wait for the installer to exit and then return cleanly.
+	doneCh := make(chan struct{})
+	go func() {
+		if err := cmd.Wait(); err != nil {
+			logger.Warnw("error while waiting for --install subprocess to complete", "err", err)
+		}
+		close(doneCh)
+	}()
```
cheukt, line 62: I don't love that we're starting a goroutine that we know will be leaked sometimes, but also don't know of a better pattern around this specific scenario. In particular, sometimes the agent process expects the `viam-agent --install` to exit cleanly and sometimes we don't.

and if we don't, the subprocess has to survive in order to re-bootstrap the new agent plist.

agent.go
```diff
@@ -27,26 +26,6 @@ type systemManager interface {
 	Enable(ctx context.Context, serviceName string) error
 }
 
-// InstallNewVersion runs the newly downloaded binary's Install() for installation of systemd files and the like.
-func InstallNewVersion(ctx context.Context, logger logging.Logger) (bool, error) {
```
cheukt, line 31: figured now that every os deals with this differently I should just split the function into the existing os specific files

cheukt (commented):
did test on both mac and linux, more extensively on macs

agent_darwin.go
```diff
@@ -10,3 +20,63 @@ const (
...
 //go:embed com.viam.agent.plist
 var serviceFileContents []byte
+
+// InstallNewVersion runs the newly downloaded binary's Install() for installation of launchd service files and the like.
+func InstallNewVersion(ctx context.Context, logger logging.Logger) (bool, error) {
+	expectedPath := filepath.Join(utils.ViamDirs.Bin, SubsystemName)
+
+	// Run the newly updated version to install launchd service files.
+	logger.Info("running viam-agent --install for new version")
+	// On macOS, --install may triggers a launchd bootout which kills the running agent's
```
EvanDorsky: mega nit: "may trigger"

agent_darwin.go
```diff
@@ -10,3 +20,63 @@ const (
...
+	// process group before the subprocess can finish bootstrapping and kickstarting the
+	// new agent. So start the subprocess in its own process group.
+	//nolint:gosec
+	cleanup := rutils.SlowLogger(
+		ctx,
+		"Waiting for new version of viam-agent to finish installing", "", "",
+		logger,
+	)
+	defer cleanup()
+	ctx, cancel := context.WithTimeout(ctx, 4*time.Minute)
```
EvanDorsky: nit: without context, 4 minutes sounds pretty long, is it worth adding a comment explaining why it's this long?

EvanDorsky (approved):
Some nits but lgtm

agent_darwin.go
jmatth: I'm a bit confused by the code comment. If the current process has been killed the goroutine can't be orphaned because the entire process is dead. I think it must be either

1. The process receives a signal it can catch like `SIGQUIT` and tries to shutdown but `main` blocks on this goroutine completing. I wouldn't consider that orphaned because the main goroutine still knows about the running goroutine
2. The process receives `SIGKILL` or a signal it can catch but main doesn't block waiting on this goroutine; the process dies and the goroutine no longer exitsts. The child process's parent changes to the parent of the dead process, which I think in this case is init. In this case it would be more accurate to say the process is orphaned (and then adopted by init)

I'm guessing it's actually `#2`

jmatth (approved):

agent_darwin.go
cheukt: yep - 2 is correct. Updated the comment to reflect that

subsystems/viamserver/viamserver.go
```diff
@@ -91,6 +91,14 @@ func (s *Subsystem) Start(ctx context.Context) error {
 	s.startStopMu.Lock()
 	defer s.startStopMu.Unlock()
 
+	// Re-check ctx after acquiring the lock: Stop() may have held it for a long time
+	// (waiting for viam-server to exit), and ctx may have been canceled in the interim.
+	// Without this check, a Start() call blocked on startStopMu would spawn an orphan
+	// viam-server process and return immediately, leaving it untracked.
```
benjirewis, line 97: I'm surprised we haven't seen this issue on Linux, too.

agent_darwin.go
benjirewis: I think what Claude generated is pretty clever. The MacOS requirement of running `launchctl bootout` instead of just `systemctl daemon-reload` more or less requires a less-than-optimal solution here. I appreciate all the documentation you + Claude added.

install.sh
```diff
@@ -284,8 +284,7 @@ main() {
 		fi
 	fi
 
-	PLIST_DEST="/Library/LaunchDaemons/com.viam.agent.plist"
-	if [ -f /etc/systemd/system/viam-agent.service ] || [ -f /usr/local/lib/systemd/system/viam-agent.service ] || [ -f "$PLIST_DEST" ]; then
+	if [ -f /etc/systemd/system/viam-agent.service ] || [ -f /usr/local/lib/systemd/system/viam-agent.service ] || ([ "$OS" = "Darwin" ] && launchctl list com.viam.agent >/dev/null 2>&1); then
```
benjirewis, line 287: Ah this is https://viam.atlassian.net/browse/APP-15754? Good catch. I suppose it's when the plist file remains but the service has been booted out.

manager.go
```diff
@@ -284,12 +284,20 @@ func (m *Manager) SubsystemUpdates(ctx context.Context) {
 		needRestartConfigChange := m.viamServer.Update(ctx, m.cfg)
 
 		if needRestart || needRestartConfigChange || m.viamServerNeedsRestart || m.viamAgentNeedsRestart {
-			if m.viamServer.RestartAllowed(ctx) {
-				m.logger.Infof("%s has allowed a restart; will restart", viamserver.SubsysName)
-				if err := m.viamServer.Stop(ctx); err != nil {
-					m.logger.Warn(err)
+			// If ctx is canceled (e.g. on macOS, launchd bootout cancels ctx before we
+			// reach this block), we know the agent is shutting down — skip RestartAllowed
+			// and stop viam-server unconditionally. Otherwise, defer to RestartAllowed.
```
benjirewis, line 289: Why do we need to _also_ `Stop` `viam-server` during subsystem updates? Shouldn't that happen as part of [`CloseAll`](https://github.com/viamrobotics/agent/blob/34ecd619d5707c8620174d75ef0729ad18fd1d97/manager.go#L486) which is invoked after the global context is canceled?

benjirewis (commented):
Thanks much for the work here and apologies for the posthumous review. Just had a couple questions/notes.

subsystems/viamserver/viamserver.go
cheukt: this specifically happened because the update also attempts to kill the agent and then the SubsystemHealthCheck starts the viam-server once more before exiting. in theory, this could happen on linux as well, but much rarer.

install.sh
cheukt: yep!

manager.go
cheukt: ya we probably could - the issue was that the context being cancelled also meant that RestartAllowed check would fail and log something confusing

benjirewis: > that RestartAllowed check would fail and log something confusing

Gotcha.

subsystems/viamserver/viamserver.go
benjirewis: Gotcha; makes sense. Thanks!

PR #223, opened 2026-04-01
RSDK-13037 - Add e2e serial tests for wifi provisioning

EvanDorsky:
Add four new tests for the wifi provisioning flow, as part of the e2e serial test suite:
- `The agent can join an unknown secure network when one is provided during wifi hotspot provisioning`
- `The agent can join an unknown insecure network when one is provided during wifi hotspot provisioning`
- `The agent can join a known secure network when one is provided during wifi hotspot provisioning`
- `Fail to connect to a network and revert to provisioning hotspot mode`

Entering provisioning mode automatically is not tested here, that's blocked by RSDK-13609 (which I want to close next).

These tests work by actually manipulating your own Mac's network interface to check for the presence of the viam provisioning hotspot and connect to it - disconnecting you from your wifi network. This makes them very annoying to run while trying to do other work, so be warned.

Note: These new tests are gated to only run on Mac for now. I'm working on Linux host compatibility but it might be a fast follow after this, depending on how ready this looks and how long Linux support takes me.

This PR also includes a few improvements to the test flow in general, like making `EnsureOnline` more flexible + robust, and adding an after-scenario hook to ensure that the agent is online after every scenario, since these tests sometimes leave the agent disconnected from the internet.

These tests run as part of the existing serial test suite: `mise r test-e2e-serial`.


internal/serialcontrol/serialcontrol.go
```diff
@@ -486,40 +500,105 @@ func (c *Client) GetDeviceArch() mo.Result[string] {
...
+		return clearRes.Error()
 	}
 
-	// Rerun the ping test. If this fails we have no way to recover
-	return c.getPingPacketLoss().FlatMap(func(value int) mo.Result[int] {
+	var connectErr error
+	for i := range 5 {
+		if i > 0 {
+			time.Sleep(time.Second * 3)
+		}
```
cheukt: why not just sleep at the bottom of the loop?

internal/serialcontrol/serialcontrol.go
```diff
@@ -486,40 +500,105 @@ func (c *Client) GetDeviceArch() mo.Result[string] {
...
+			connectErr = fmt.Errorf("nmcli connection up failed: %s", upOutput)
+		} else {
+			connectErr = fmt.Errorf("nmcli connect failed: %s", output)
+		}
+	}
+	if connectErr != nil {
+		return errw.Wrap(connectErr, "failed to connect to wifi network after retries")
+	}
+
+	time.Sleep(time.Second * 2)
```
cheukt, line 570: what is this sleep for?

internal/serialcontrol/serialcontrol.go
```diff
@@ -486,40 +500,105 @@ func (c *Client) GetDeviceArch() mo.Result[string] {
...
 			return mo.Err[int](fmt.Errorf("internet connection unstable with packet loss of %d%%", value))
 		}
 		return mo.Ok(value)
 	}).Error()
 }
 
+// CanPing attempts to ping app.viam.com and returns false
+// the ping returns a nonzero status code, true if it returns "0",
+// and error if the command doesn't run properly
+func (c *Client) CanPing() mo.Result[bool] {
```
cheukt: is getPingPacketLoss not enough here? feels redundant to run the same command twice

features/serial/provision-wifi.feature
```diff
@@ -0,0 +1,32 @@
...
+    And the tester shares a secure wifi network
+    Then the provisioning hotspot goes away
+    And viam-agent can reach the app
+  Scenario: The agent can join a known secure network when one is provided during wifi hotspot provisioning
+    When viam-agent is in forced provisioning mode
+    And the provisioning hotspot comes up
+    And the tester shares a secure wifi network
+    Then the provisioning hotspot goes away
+    And viam-agent can reach the app
+  # Scenario: Connect to an insecure network
```
cheukt: what about this scenario?

features/serial/provision-wifi.feature
```diff
@@ -0,0 +1,32 @@
+@darwin
+Feature: wifi provisioning
+  Background:
+    Given viam-agent is installed
+    And the viam-agent systemd unit is enabled
+    And the viam-agent systemd unit is running
+  # Scenario: The agent enables the provisioning hotspot when the internet connection is lost
```
cheukt: let's ticket these scenarios out instead of commenting them out

agent_serial_test.go
```diff
@@ -124,8 +133,21 @@ func InitializeSuite(t *testing.T) func(*godog.TestSuiteContext) {
 			}
 		})
 		tsc.AfterSuite(func() {
-			ctx, cancel := context.WithTimeout(context.Background(), time.Second*10)
+			// Give a pretty long timeout since there's actual user interaction in the loop now
```
cheukt: i don't think i see the user interaction here, am i missing something?

agent_serial_test.go
```diff
@@ -358,6 +407,98 @@ func testViamFilesRemoved(ctx context.Context) (context.Context, error) {
...
+	return ctx, nil
+}
+
+func testProvisioningHotspotDisables(ctx context.Context) (context.Context, error) {
+	// this script checks for the provisioning network every 5 seconds, then connects to it
+	// if it finds it
+	cmd := exec.Command("bash", "cmd/test-client/test_provisioning_network_gone.sh")
+	cmd.Env = append(os.Environ(), "HOSTNAME="+hostName)
+
+	// TODO: this should error out if cmd returns a nonzero exit code
```
cheukt: can you clean up the todos and make a ticket if you don't plan on addressing in this pr?

cheukt (commented):
looks good so far, some questions

internal/serialcontrol/serialcontrol.go
EvanDorsky: getPingPacketLoss would panic on the "expected match in %s" case sometimes when the pi wasn't connected to the internet, so I decided to first check for an internet connection before checking packet loss. I thought about adding this check within getPingPacketLoss itself, but then it would have to signal no internet by reporting 100% packet loss... but now looking at the code I can see that other failure cases are already handled this way. I guess it would be cleaner to make getPingPacketLoss robust to the case causing the panic, and then CanPing won't be necessary.

features/serial/provision-wifi.feature
EvanDorsky: Good call I forgot about this one, should be easy to add it in.

agent_serial_test.go
EvanDorsky: Yeah this comment is out of date from an earlier version where I actually prompted the user to join a wifi network, I'll remove it

EvanDorsky:
I've added the insecure wifi test scenario, ticketed out the remaining work, fixed some issues and addressed the comments, so this is ready for re-review.

cheukt (approved):
ran it and passed - LGTM but would wait for another pair of eyes on this too

internal/serialcontrol/serialcontrol.go
```diff
@@ -486,40 +496,103 @@ func (c *Client) GetDeviceArch() mo.Result[string] {
...
+			packetLoss,
 		)
 	}
 
-	// If we're offline try connecting to the specified wifi network.
-	connectWifiRes := c.runCmd(fmt.Sprintf(`nmcli device wifi connect "%s" password "%s"`, ssid, password))
-	if connectWifiRes.IsError() {
-		return errw.Wrap(connectWifiRes.Error(), "failed to connect to wifi network")
+	if ssid == "" || password == "" {
+		return fmt.Errorf("device offline with bad ping but no wifi credentials provided, cannot continue")
```
danielbotros, line 532: Copy paste accident? (two checks for `ssid == "" || passowrd == ""`)

cmd/test-client/test_provisioning_join_network.sh
```diff
@@ -0,0 +1,17 @@
...
+    result=$(networksetup -setairportnetwork en0 "$network" "viamsetup" 2>&1)
+    if [ -z "$result" ]; then
+        echo "connected to $network"
+        # sleep a bit even on success so the next step doesn't try too early
+        sleep 5
+        exit 0
+    fi
+    sleep 5
+done
+
```
danielbotros, line 16: nit: Can we echo a failure reason here like the other scripts?

internal/serialcontrol/serialcontrol.go
```diff
@@ -486,40 +496,103 @@ func (c *Client) GetDeviceArch() mo.Result[string] {
...
+			if strings.Contains(upOutput, "successfully activated") {
+				connectErr = nil
+				break
+			}
+			// If that fails, try "%s@wlan0", since viam-agent adds that suffix to connections it creates.
+			upRes = c.runCmd(fmt.Sprintf(`nmcli connection up "%s@wlan0"`, ssid))
+			upOutput = strings.Join(upRes.OrEmpty(), " ")
+			if strings.Contains(upOutput, "successfully activated") {
+				connectErr = nil
+				break
```
danielbotros, line 557: This block seems a little hacky that we try to attempt both SSIDs. If we know the network is already saved is there a way we can query the system for it's actual name and only try that instead?

internal/serialcontrol/serialcontrol.go
danielbotros: Might be missing some context but I agree at a high level that if we can fix `getPingPacketLoss` to work generally we should drop `CanPing`

agent_serial_test.go
```diff
@@ -124,8 +138,18 @@ func InitializeSuite(t *testing.T) func(*godog.TestSuiteContext) {
...
+			ctx, cancel := context.WithTimeout(context.Background(), time.Second*20)
 			defer cancel()
+			if err := serialClient.EnsureOnline(cfg.Wifi.SSID, cfg.Wifi.Password); err != nil {
+				t.Logf("error reconnecting to wifi during cleanup: %v", err)
+			}
+			if runtime.GOOS == "darwin" {
+				if _, err := hostEnsureOnline(ctx); err != nil {
+					t.Logf("error restoring host connection to internet during cleanup: %v", err)
+				}
+			}
```
danielbotros, line 150: If we call this after every step, do we also need it in `AfterSuite`? Think it's fine to just leave it here too anyway just curious

internal/serialcontrol/serialcontrol.go
```diff
@@ -272,6 +272,16 @@ func (c *Client) Sudo() error {
...
 
+func (c *Client) ClearWifiConnections() mo.Result[[]string] {
+	// clear all wifi connections except for the setup hotspot
+	return c.runCmd(`nmcli -f NAME,TYPE connection show | awk '$NF == "wifi" && !/viam-setup/ {NF--; print}'` +
+		` | xargs -I{} nmcli connection delete "{}"`)
+}
+
+func (c *Client) ListWifiConnections() mo.Result[[]string] {
+	return c.runCmd(`nmcli -f NAME,TYPE connection show | awk '$NF == "wifi" && !/viam-setup/ {NF--; print}'`)
+}
```
danielbotros, line 283: Since these are almost the same command can we extract the duplicated part out into its own string so that it doesn't have to be changed in two places in the future?

features/serial/provision-wifi.feature
```diff
@@ -0,0 +1,36 @@
+@darwin
+Feature: wifi provisioning
+  Background:
+    Given viam-agent is installed
+    And the viam-agent systemd unit is enabled
+    And the viam-agent systemd unit is running
+  Scenario: The agent can join an unknown insecure network when one is provided during wifi hotspot provisioning
+    When there are no available wifi networks
```
danielbotros, line 8: mega nit: I don't think this changes test behavior but should `there are no available wifi networks`  be a `Given` instead of `When` statement since it sets up a test precondition?

agent_serial_test.go
```diff
@@ -358,6 +415,114 @@ func testViamFilesRemoved(ctx context.Context) (context.Context, error) {
...
+		if res.IsError() {
+			return ctx, fmt.Errorf("canPing failed: %w", res.Error())
+		}
+		if !res.MustGet() {
+			return ctx, nil
+		}
+		time.Sleep(5 * time.Second)
+	}
+	return ctx, fmt.Errorf("viam-agent did not go offline within timeout")
+}
```
danielbotros, line 482: 2.5 minute timeout seems long for this, test suite already takes a while to run

danielbotros (commented):
Ran tests and passed. Super awesome addition to the suite, thanks Evan! Left some minor style comments.

PR #225, opened 2026-04-13
RSDK-13736 RSDK-13509 e2e tests for automatic provisioning and live systemd settings

EvanDorsky:
This branch adds a new test for the automatic provisioning flow (ensuring that the provisioning hotspot comes up when the connection to the app is lost) (RSDK-13736), and also adds tests ensuring that the systemd setting changes made by agent actually go live during the test (RSDK-13509).

This is intended as a fast follow to https://github.com/viamrobotics/agent/pull/223 and is a much smaller branch than that one. https://github.com/viamrobotics/agent/pull/223 should be merged first.

This branch also includes improvements to how wifi network joining / scanning is done from the host side. I pulled the retry loop into golang, which improves reliability and edge case handling while making it easier to add linux host support in the future.

Also includes a few improvements suggested in https://github.com/viamrobotics/agent/pull/223


internal/serialcontrol/serialcontrol.go
danielbotros: Hmm, since we loop attempts and can error multiple times and in different ways can we maybe make `connectErr` a `multierr` and collect and return them all here?

internal/serialcontrol/serialcontrol.go
```diff
@@ -528,38 +538,44 @@ func (c *Client) EnsureOnline(ssid, password string) error {
...
-			upRes = c.runCmd(fmt.Sprintf(`nmcli connection up "%s@wlan0"`, ssid))
-			upOutput = strings.Join(upRes.OrEmpty(), " ")
-			if strings.Contains(upOutput, "successfully activated") {
-				connectErr = nil
-				break
-			}
-			connectErr = fmt.Errorf("nmcli connection up failed: %s", upOutput)
-		} else {
-			connectErr = fmt.Errorf("nmcli connect failed: %s", output)
+			connectErr = fmt.Errorf("nmcli connection up %q failed: %s", connName, upOutput)
```
danielbotros: Will the `OrEmpty()` swallow errors? If so, I don't think the `upOutput` log field here adds anything. Same with `connectWifiRes`

agent_serial_test.go
```diff
@@ -415,6 +418,62 @@ func testViamFilesRemoved(ctx context.Context) (context.Context, error) {
...
+	var err error
+	for i := range 30 {
+		if i > 0 {
+			time.Sleep(time.Second * 2)
+		}
+		// Check the most recent journald startup log line which reports the
+		// live max size and journal path. Persistent storage uses
+		// /var/log/journal/ (volatile would be /run/log/journal/).
+		res := serialClient.RunScript(
+			`journalctl --no-pager -u systemd-journald -n 5 --output=short-monotonic 2>&1 | grep "System Journal"`,
```
danielbotros: Could only 5 logs cause this to flake?

agent_serial_test.go
```diff
@@ -415,6 +418,62 @@ func testViamFilesRemoved(ctx context.Context) (context.Context, error) {
...
+		)
+		if res.IsError() {
+			err = res.Error()
+			continue
+		}
+		output := strings.Join(res.MustGet(), "")
+		if !strings.Contains(output, "EXISTS") {
+			err = fmt.Errorf("wifi power save config not found at %s", wifiPowerSaveFilepath)
+			continue
+		}
```
danielbotros, line 479: If we find the file we return `ctx, nil`, if we don't then I think `res.IsError()` branch catches it and so this isn't reachable I believe

agent_serial_test.go
```diff
@@ -424,75 +483,139 @@ func testClearWifiConnections(ctx context.Context) (context.Context, error) {
...
+
+	var lastOut string
+	for time.Now().Before(startTime.Add(provTimeout)) {
+		cmd := exec.Command("networksetup", "-setairportnetwork", "en0", hotspotName, "viamsetup")
+
+		out, err := cmd.CombinedOutput()
+		outStr := string(out)
+		// this rarely happens, err is only non-nil when the cmd fails to run at all
+		if err != nil {
+			return ctx, fmt.Errorf("joining provisioning hotspot failed: %w\n%s", err, out)
```
danielbotros, line 522: Should we `continue` and try again or does the command failing to run mean there is a larger problem we should abort the test early on? Same to the early returns below

danielbotros (commented):
Have not run myself yet but overall LGTM, some minor comments / questions

cheukt (approved):
generally lgtm % daniels comments. ran the suite and it passed

internal/serialcontrol/serialcontrol.go
EvanDorsky: Ah, yeah, I don't think I want to use OrEmpty. Thanks for catching that.

EvanDorsky: Actually I think this is mostly okay, I should just catch any error before using OrEmpty. I think the idea here is that if the command successfully runs but fails to join the network, upOutput will contain the reason printed by nmcli (like network not present, failed to add, etc.). There will only be an actual error if the command fails to run at all, like from a polluted tty or a serial failure.

agent_serial_test.go
EvanDorsky: Looking at the logs, `systemd-journald` is pretty quiet so I'm not too worried about that, I'm actually more worried about it just finding logs from a previous run since it has no way to discriminate. I'm going to use `journalctl --header` instead of reading log lines, seems less risky.

EvanDorsky: Ah actually - I can get the pid of the actively running `systemd-journald` process and then match it to the log lines. Then I can also check a much larger number of log lines and not worry about catching old logs.

agent_serial_test.go
EvanDorsky: I think this is the same thing as above where, unfortunately, RunScript doesn't return error if the return code of the script is nonzero - so we need to manually handle things this way.

agent_serial_test.go
EvanDorsky: Good question. I've never seen a test continue in a bad state after returning early from one of these - I think that if something has gone wrong with the connection, the after scenario / after suite hooks will catch it and panic. I also kind of want to keep any potential panics in those hooks wherever possible so that we don't have to think about whether any given error in a test step should cause a panic.

danielbotros (approved):

PR #226, opened 2026-04-15
RSDK-13038 BLE provisioning tests

EvanDorsky:
I made a few small changes to the existing provisioning client to get BLE provisioning working, and wrote a new suite for testing it, with one scenario for now.

This is only tested on mac for now, but it should work on linux - would be great if someone who uses linux @jmatth could try running this suite.

In order to get linux support for the test, I had to split the provisioning client into its own go module, because the test requires changes to https://github.com/tinygo-org/bluetooth that aren't in a tagged release yet (and in any case I didn't want to bump the version of https://github.com/tinygo-org/bluetooth just for the tests).

Splitting the provisioning client seemed like a pretty lightweight and isolated change - I mostly had claude do it - but definitely let me know if this was either a bad idea or implemented in a bad way. I re-ran the wifi provisioning tests after doing this and they still passed.


cmd/provisioning-client/bluetooth.go
```diff
@@ -125,7 +125,7 @@ func ExitProvisioning(chars map[string]bluetooth.DeviceCharacteristic) error {
 		return err
 	}
 
-	_, err = chars[exitProvisioningKey].WriteWithoutResponse(cryptExit)
```
cheukt, line 128: what's the difference between `Write` and `WriteWithoutResponse`?

cmd/provisioning-client/provisioning-client
cheukt, line 1: definitely remove this binary before merge

features/serial/install.feature
```diff
@@ -5,3 +5,5 @@ Feature: install viam-agent
     When viam-agent is installed
     Then the viam-agent systemd unit is running
     And the viam-agent systemd unit is enabled
+    And the journald config is live
+    And the wifi power save config is live
```
cheukt, line 9: can you rebase from main/clean up merge conflicts so the diff is cleaner?

cmd/provisioning-client/go.mod
```diff
@@ -0,0 +1,32 @@
+module github.com/viamrobotics/agent/cmd/provisioning-client
```
cheukt: I'd like to avoid adding the go.mod and go.sum to the repo, we should be able to do a `go run ./cmd/provisioning-client` and use the go.mod from the agent repo

agent_serial_test.go
```diff
@@ -519,6 +646,60 @@ func testSendSecureConnectionInfo(ctx context.Context) (context.Context, error)
...
+	for range 3 {
+		if lastErr != nil {
+			time.Sleep(1 * time.Second)
+		}
+		cmd := exec.CommandContext(ctx, "go", "run", ".",
+			"-b",
+			"--status",
+			"--filter", filter,
+		)
+		cmd.Dir = "cmd/provisioning-client"
```
cheukt: from above 

> I'd like to avoid adding the go.mod and go.sum to the repo, we should be able to do a `go run ./cmd/provisioning-client` and use the go.mod from the agent repo

would that work with this setup?

cheukt (changes requested):
looks good! but going to wait to do the full review after the merge conflicts are cleaned up and the diff is cleaner

cheukt:
I did run the tests and it passed, and the general flow makes sense to me

cmd/provisioning-client/bluetooth.go
EvanDorsky: `Write` automatically supports messages longer than 23 bytes (the default BLE MTU) by splitting long messages into a series of smaller ones, while `WriteWithoutResponse` doesn’t. I looked into whether it was possible to change the MTU so I wouldn’t have to change the method but the library didn’t have any support for that. Writes weren’t working at all until I switched to `Write`, which made sense to me because all the messages we were trying to send were longer than 23 bytes.

features/serial/install.feature
EvanDorsky: Yes, sorry about that - now that the other PR is merged I can rebase, I’ll do that soon.

cmd/provisioning-client/go.mod
EvanDorsky: I did that for this reason

> In order to get linux support for the test, I had to split the provisioning client into its own go module, because the test requires changes to https://github.com/tinygo-org/bluetooth that aren't in a tagged release yet

I didn’t want to make a new module either, but doing that seemed like the cleanest way to bump the bluetooth library to support the test without bumping agent’s bluetooth library to an unreleased version. Also Linux CI actually fails if I don’t do this - I think the build fails because `Write` isn’t in agent’s version of the bluetooth library on Linux. So I’d have to somehow fix the Linux build if we opted against the separate module.

Open to another approach though, let me know what you think.

cheukt: [tinygo](https://github.com/tinygo-org/bluetooth) they just released a new version, can you test it? i would rather do the upgrade for the whole repo than separately maintain deps

EvanDorsky: Oh wow nice they literally just did the release 3 hours ago - yeah I'll test it!

EvanDorsky: image

Incredible

cmd/provisioning-client/provisioning-client
EvanDorsky: Oops. Yeah, I'm going to force push a new branch soon off of current main, which will also remove the binary, since I accidentally committed the binary while splitting out the provisioning client into its own module.

EvanDorsky:
Finally pushed a much cleaner diff, took a while because I was running into some local setup issues while testing. Still need to evaluate agent with the new bluetooth library version.

danielbotros (approved):
LGTM! Much needed test

jmatth (commented):
I tried running this on my CachyOS install and couldn't get it to succeed. The provisioning client would fail at varying points in the process. It seemed to get further if I turned off my bluetooth mouse and toggled bluetooth on my machine first. If this works on macOS I'm not sure it's worth the time to track down what's broken on Linux. The one action item I think that should come out of my testing is changing how `provisioning-client` handles failure. Right now it will print error logs but when the process eventually exits it does so with `0`, so the E2E test harness thinks it succeeded and continues to the next step. This doesn't create a false positive since the test eventually fails at a later step, but it points the engineer toward the wrong part of the codebase to start investigating the failure. Depending on how much effort it is I think we should

- [ ] Modify `provisioning-client` to return a non-zero exit code or otherwise communicate failure to the test harness
- [ ] Update the steps that run `provisioning-client` to log the exact command they're running at the `INFO` level and any logs the command produces at the `DEBUG` level, which matches the log levels used for commands run over the serial connection

For reference, this is the closest `provisioning-client` ever got to succeeding:

```
%> go run ./cmd/provisioning-client -b --filter viam-setup-jmatth-deskpi ...<further params redacted>

Scanning...
Found device: viam-setup-jmatth-deskpi [D8:3A:DD:D7:8B:51]
Connecting...
Discovering characteristics for service UUID: 51d984c2-43bd-5228-a3ac-a3eca992e8b3
Found: cd5b8fb9-4006-56e5-a78b-044cf6ae48cb (psk)
Found: 10cd57f1-01bb-5937-89fb-64cc40be53d2 (exit_provisioning)
Found: 6e164616-ee96-5835-b199-115ddbcd885f (agent_version)
Found: d0029d11-a8c2-5231-8fcc-83de66952f01 (secret)
Found: f30e26d6-155c-5a24-bd7f-6b1a40e463da (api_key)
Found: 37a720ee-86e5-55a8-b876-d200fb7e4f72 (ssid)
Found: 444ee2b0-b3fa-5d74-bb35-f194642188b3 (app_address)
Found: 61eba4df-b901-502c-b278-fadf9d52802b (networks)
Found: 70bc8310-68ca-5011-9282-72f201293dfb (pub_key)
Found: 96a4bebb-a361-5c73-9d76-45a0faa9d4a0 (unlock_pairing)
Found: c2be234e-1975-5e85-b97b-bb30c3bf43d2 (id)
Found: ea8a8689-548f-5941-829b-82aeff8095b7 (status)
Found: 49d34f00-cf76-55fa-9f8d-23f6836136ab (manufacturer)
Found: 75f64044-b604-5547-a723-7f8618255ddc (errors)
Found: e7e1ac15-fb59-54ab-8538-72008ad4ee43 (model)
Found: d2eae9e8-30bc-5fdf-bd9f-bd4e8a4017d2 (fragment_id)
Writing device credentials...
2026-04-28T20:15:19.196Z        ERROR   provisioning-client     provisioning-client/main.go:15  writing part id: Operation failed with ATT error: 0x0e
```

EvanDorsky:
Thanks for testing @jmatth. Independent of the error code behavior of `provisioning-client`, I did find a place where my test code was swallowing some errors (golang variable shadowing, a fun footgun I learned about yesterday) and causing a step to pass when it should have failed. You might have been running into that while testing as well.

Aside from that, I agree it makes sense to stick to mac for these tests for now.

> Update the steps that run provisioning-client to log the exact command they're running at the INFO level and any logs the command produces at the DEBUG level, which matches the log levels used for commands run over the serial connection

This is a good call, I'll do this.

> Modify provisioning-client to return a non-zero exit code or otherwise communicate failure to the test harness

I'll look into this too. The new commits actually do a much better job catching errors from `provisioning-client` by just parsing the output (which is already necessary to verify that a given behavior raises the correct error message).

EvanDorsky:
New changes:
- added test cases for automatic provisioning startup, known/unknown secure networks, insecure network, and bad wifi passkey. Basically parity with the wifi tests.
- added expected bluetooth characteristic checking to the "bluetooth is discoverable" case (this eliminates false positives on incomplete connections, which do happen)
- added error string parsing for the "bad wifi passkey" case (this helped catch some bugs)
- fixed the false pass of the "bluetooth is discoverable" step that @jmatth pointed out by catching errors properly
- added info / debug logs to the bluetooth steps (which required promoting the test logger to a package-level variable)
- removed the call to serialcontrol `EnsureOnline` from the after suite hook since it was basically just thrashing network manager (since the test background steps delete all connections anyway)

Of particular note: I removed the "invalid SSID" case for two reasons
1. I don't think a real client can actually provide an SSID that agent doesn't know, because the client gets the candidate list from agent
2. Forcing this case with the command-line provisioning client causes an indefinite hang (even on main) and needs further investigation

I tested the bluetooth suite (the tests are still darwin only), and also tested provisioning with the actual iPhone and Android apps (since the next release will contain the version bump to the bluetooth library) and everything looks good, but could still use some more runs on the test suite. I also still need to make sure the full suite still passes, didn't have time to run that today - but I think this is ready for re-review in the meantime.

.vscode/settings.json
```diff
@@ -0,0 +1,5 @@
+{
```
cheukt: should this be committed?

agent-test.work.toml
```diff
@@ -0,0 +1,24 @@
+# Uncommented fields are required. Text <in brackets> is a placeholder and
```
cheukt: this also shouldn't be committed

report.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
cheukt: should this be merged too?

cheukt (commented):
looks good, probably just need to delete a few files. will approve once it is clean

.vscode/settings.json
EvanDorsky: nope - I'm just noticing I accidentally committed a couple of files, will be removing them

agent-test.work.toml
EvanDorsky: yeah fixing this now my bad

EvanDorsky:
apologies for the bad file push, those files have been removed from the history

cheukt (approved):
LGTM!

PR #227, opened 2026-04-16
[RSDK-13934] Decouple BLE lifecycle from WiFi provisioning

ale7714:
## Summary

BLE advertising and WiFi hotspot previously shared a single lifecycle — both started in `startProvisioning()` and stopped in `stopProvisioning()`, gated on the same timers (`OfflineBeforeStartingHotspotMinutes`, `UserIdleMinutes`). Those timers are calibrated for WiFi hotspot (costly to false-activate), but BLE advertising runs in parallel, disrupts nothing, and has no radio to reclaim. 

## Changes

- **`bleTracker` struct** couples BLE lifecycle state and the advertisement handle so they're always updated together. Three purpose-built methods enforce invariants at the mutation site:
  - `setStarting()` — transitions to `bleStarting`, no adv involvement
  - `setRunning()` — transitions to `bleRunning`, errors if adv is nil
  - `clearAndSetOff()` — stops adv, nils it, transitions to `bleOff`
  
  There is no way to reach `bleOff` without clearing the advertisement, and no way to reach `bleRunning` without one.
- **`bleLoop` goroutine** (1s tick) owns all BLE state. It runs a reconciler (`reconcileBluetooth`) that compares desired vs. actual state and calls the existing start/stop primitives. Characteristic values are pushed from `bleLoop`, not `mainLoop`.
- **`startProvisioning`/`stopProvisioning`** no longer touch BLE — WiFi hotspot only.
- **`noBT` sticky flag removed.** A single BlueZ failure previously disabled BLE for the entire process lifetime. Bounded exponential backoff (5s doubling to 10m) replaces it: transient failures self-heal, and once backoff reaches the cap retries stop with a log suggesting `disable_bt_provisioning` if the device lacks bluetooth.
- **`decideBleAction`** is a pure function (no receiver, no I/O) for testability.
- **Visible networks cache** prevents `getVisibleNetworks()` from being recomputed every bleLoop tick. `backgroundLoop` refreshes the cache after each scan.

## Future work

The reconciler currently polls on a 1s ticker. A natural next step would be replacing the ticker with an event-driven channel so that `mainLoop` (or any other goroutine) can signal `bleLoop` to re-evaluate immediately — e.g. when connectivity changes or config is updated — instead of waiting up to a second for the next tick. This PR doesn't prevent that: `bleLoop`'s `select` can swap `tick.C` for a channel receive with no changes to `bleTracker`, `decideBleAction`, or the reconciler itself. That's a larger refactor (it touches the mainLoop/backgroundLoop signaling too) so it's left for a follow-up.


cheukt:
i like the desired state behavior - will probably make it easier to reason about. changes all make sense too

ale7714:
@cheukt if directionally this makes sense, i'll start testing and request reviews again

cheukt:
ya I think so!

subsystems/networking/bluetooth_linux.go
```diff
@@ -17,31 +19,78 @@ const (
...
+const (
+	bleOff      bleState = iota // not started, or cleanly stopped
+	bleStarting                 // partial init in progress, Start() not yet succeeded
+	bleRunning                  // Start() succeeded, advertising
 )
 
-// startProvisioningBluetooth should only be called by 'StartProvisioning' (to ensure opMutex is acquired).
+// bleTracker couples BLE state and advertisement so they're always updated together.
+type bleTracker struct {
+	state atomic.Int32
```
abe-winter: maybe comment that this is a `bleState`

subsystems/networking/bluetooth_linux.go
```diff
@@ -17,31 +19,78 @@ const (
...
+	adv   *bluetooth.Advertisement // only accessed from bleLoop
+}
+
+func (bt *bleTracker) getState() bleState {
+	return bleState(bt.state.Load())
+}
+
+func (bt *bleTracker) setStarting() {
+	bt.state.Store(int32(bleStarting))
```
abe-winter, line 64: warn when transition from running to starting?

subsystems/networking/bluetooth_linux.go
```diff
@@ -54,35 +103,49 @@ func (n *Subsystem) startProvisioningBluetooth(ctx context.Context) error {
...
 	}
 
 	// Start advertising the bluetooth service.
-	if err := n.btAdv.Start(); err != nil {
+	if err := n.ble.adv.Start(); err != nil {
+		n.cleanupPartialBluetooth()
 		return fmt.Errorf("failed to start advertising: %w", err)
 	}
-	n.btHealthy = true
+	if err := n.ble.setRunning(); err != nil {
```
abe-winter: if Start and setRunning are always called together, should they be a single `n.ble.start()` call?

subsystems/networking/bluetooth_linux.go
```diff
@@ -54,35 +103,49 @@ func (n *Subsystem) startProvisioningBluetooth(ctx context.Context) error {
...
+		n.cleanupPartialBluetooth()
+		return err
+	}
 
 	n.logger.Info("Bluetooth provisioning started.")
 	return nil
 }
 
-// stop stops advertising a bluetooth service which (when enabled) accepts WiFi and Viam cloud config credentials.
+// cleanupPartialBluetooth is idempotent BLE teardown; transitions to bleOff.
```
abe-winter: idempotent in our codebase, but is this true on the system as well? (like can bluez get stuck? if yes do we need a 'stopping' state?)

subsystems/networking/bluetooth_linux.go
```diff
@@ -54,35 +103,49 @@ func (n *Subsystem) startProvisioningBluetooth(ctx context.Context) error {
...
+func (n *Subsystem) cleanupPartialBluetooth() {
+	n.ble.clearAndSetOff(n.logger)
+	if err := n.disablePairing(); err != nil {
+		n.logger.Warnw("error disabling BT pairing during rollback", "err", err)
+	}
+	if err := n.removeServices(); err != nil {
+		n.logger.Debugw("no gatt services to remove during rollback", "err", err)
```
abe-winter: are you sure this error will always be 'no gatt services'?

subsystems/networking/bluetooth_linux.go
```diff
@@ -54,35 +103,49 @@ func (n *Subsystem) startProvisioningBluetooth(ctx context.Context) error {
...
+	}
+}
+
+// stopProvisioningBluetooth stops BLE advertising and tears down the gatt service.
+// Must only be called from bleLoop.
 func (n *Subsystem) stopProvisioningBluetooth() error {
-	if n.btAdv == nil {
+	currentState := n.ble.getState()
+	if currentState == bleOff {
```
abe-winter: could use switch here; that allows us to lint for whether the enum is fully handled

subsystems/networking/bluetooth_linux.go
abe-winter, line 167: out of scope for your review, but at some point we should go through the codebase and clearly distinguish between ble + classic. 'bluetooth' by itself is ambiguous

subsystems/networking/bluetooth_loop_linux.go
```diff
@@ -0,0 +1,179 @@
...
+const bleLoopTick = time.Second
+
+// bleLoop owns BLE lifecycle and characteristic updates. Single writer of all BLE state.
+func (n *Subsystem) bleLoop(ctx context.Context) {
+	defer utils.Recover(n.logger, nil)
+	defer n.monitorWorkers.Done()
+	defer func() {
+		if err := n.stopProvisioningBluetooth(); err != nil {
+			n.logger.Warnw(
+				"ble_reconcile_stop_failed",
```
abe-winter: might add a human readable message

subsystems/networking/bluetooth_loop_linux.go
```diff
@@ -0,0 +1,179 @@
...
+func (n *Subsystem) pushBluetoothCharacteristics() {
+	isOnline := n.connState.getOnline()
+	isConnected := n.connState.getConnected()
+	hasConnectivity := isConnected || isOnline
+	if n.Config().TurnOnHotspotIfWifiHasNoInternet.Get() {
+		hasConnectivity = isOnline
+	}
+	isConfigured := n.connState.getConfigured()
+
+	if err := n.btChar.updateStatus(isConfigured, hasConnectivity); err != nil {
```
abe-winter: is there any downside to writing these every time, rather than only on change?

subsystems/networking/bluetooth_loop_linux.go
```diff
@@ -0,0 +1,179 @@
...
+}
+
+// reconcileBluetooth converges BLE state toward the desired state. Called from bleLoop only.
+func (n *Subsystem) reconcileBluetooth(ctx context.Context) {
+	action := decideBleAction(bleDecisionInput{
+		desired:          n.bluetoothDesired(),
+		currentState:     n.ble.getState(),
+		now:              time.Now(),
+		nextAttempt:      n.bleNextAttempt,
+		retriesExhausted: n.bleBackoff >= bleBackoffMax,
```
abe-winter: 'max' implies 'max allowed'. is this saying that 'max = exhausted'?

if this starts at 5 minutes + doubles, there isn't a backoff behavior

subsystems/networking/bluetooth_loop_linux.go
```diff
@@ -0,0 +1,179 @@
...
+type bleAction int
+
+const (
+	bleActionNone bleAction = iota
+	bleActionStart
+	bleActionStop
+)
+
+type bleDecisionInput struct {
+	desired          bool
```
abe-winter: maybe comment this to make clear it's 'do we want bluetooth running', rather than 'desired state'. or rename 'desiredRunning'. reading this, at first I thought it was 'desired state' (vs current state).

subsystems/networking/bluetooth_loop_linux.go
```diff
@@ -0,0 +1,179 @@
...
+			)
+			if n.bleBackoff >= bleBackoffMax {
+				n.logger.Warn("BLE startup keeps failing. If this device has no bluetooth hardware, " +
+					"set disable_bt_provisioning to avoid these retries.")
+			}
+			return
+		}
+		n.bleBackoffReset()
+		n.logger.Infow(
+			"ble_reconcile_started",
```
abe-winter: does 'started' here mean 'start was called' or 'start completed and ble is running'?

subsystems/networking/networking_linux.go
```diff
@@ -261,6 +264,11 @@ func (n *Subsystem) Start(ctx context.Context) error {
 		n.mainLoopHealth.MarkGood()
 		n.bgLoopHealth.MarkGood()
 		go n.mainLoop(cancelCtx)
+
+		if !n.Config().DisableBTProvisioning.Get() {
```
abe-winter, line 278: will Config.DisableBTProvisioning ever change during a run? (i.e. if config read from cloud?)

subsystems/networking/networkmanager_linux.go
```diff
@@ -59,6 +59,19 @@ func (n *Subsystem) warnIfMultiplePrimaryNetworks() {
 	}
 }
 
+func (n *Subsystem) refreshVisibleNetworksCache() {
+	networks := n.getVisibleNetworks()
+	n.visibleNetworksMu.Lock()
+	n.visibleNetworksCache = networks
+	n.visibleNetworksMu.Unlock()
```
abe-winter: could use defer unlock pattern

abe-winter (approved):
seems good (subject to tests obv)

left a bunch of nits from close reading, feel free to ignore

how does the main networking loop signal the bleLoop now? just `ctx.Done`, `go n.bleLoop(cancelCtx)`, and the visible networks cache?

subsystems/networking/networking_linux.go
ale7714: Any change to NetworkConfiguration (which contains DisableBTProvisioning) makes Update return `needRestart=true` which makes the agent restart.

ale7714:
@abe-winter 

> how does the main networking loop signal the bleLoop now? just ctx.Done, go n.bleLoop(cancelCtx), and the visible networks cache? 

Yes:
* mainLoop, backgroundLoop, and bleLoop all share cancelCtx from Start(), so cancellation tears down all three.
* mainLoop/backgroundLoop write online/connected/configured. bleLoop reads them in bleDesired() and pushBleCharacteristics(). This is the actual desired-state input (alongside the DisableBTProvisioning config). 
* backgroundLoop refreshes visibleNetworksCache  after each scan; bleLoop reads via cachedVisibleNetworks().

ale7714:
many thanks to @EvanDorsky for running the e2e tests and fixing some issues

@cheukt i did a couple more round of testing both regular flow using the flutter example project and also the e2e tests. It seems to be working fine now.

cheukt (approved):
lgtm - didn't get a chance to run manual tests but generally looks good. let me know if there are specific test cases you want me to check out

ale7714:
@cheukt  Important observation that @EvanDorsky raised. 

With my PR, BLE no longer goes down during a wifi connect attempt. This is behavior change: a client polling BLE right after submitting bad credentials gets errors=[] for up to ~50s (NetworkManager's connect timeout) before the failure is reported, where on main branch BLE was down during that window so the client couldn't poll at all.

Evan proposed we could still tear BLE down around the connect attempt to match main's behavior as it would be the lowest effort and no impact to clients. I agree with the instinct but I'd like to try to keep the behavior on this branch. Keeping BLE on allows for smoother handling when someone enters wrong credentials. 

BLE is currently used by DY (and soon Gambit) through https://github.com/viamrobotics/viam_flutter_bluetooth_provisioning_widget.  I took a look at the widget and it doesn't rely on BLE going away during the connect attempt. After credentials are sent, it runs a 5s timer; on each tick it:

* Asks the Viam app for the machine's lastAccess — if it's within 10s, declares success and stops.
* Otherwise, if still connected over BLE, reads the errors characteristic. If non-empty, declares failure and stops.
* Otherwise, keeps polling.

Empty errors is treated as "no result yet", not as success. So on both branches the widget shows checking… → success or checking… → errorConnecting. Identical user-visible UX. This is why I didn't noticed the issue when testing with the  flutter provisioning widget.  

The diagnosed bug is real on our e2e because we tied BLE being on/off as a signal of failure or success. I'd like to propose two options going forward:

1. Keep the behavior on the branch as is. We should document the on-the-wire contract: empty errors = "no result yet", not "no error" and update the e2e test so it follows that contract as well. 

2.  Keep the behavior and add an explicit "connecting" state to BLE status (bit 2 of the status byte). It would be backward-compatible because the widget reads bits 0–1 today; a new bit 2 is invisible to existing clients. This would allow us to improve the widget in future iterations. 

I'd prefer 1 as this PR has gotten quite large. 

Thoughts?

cheukt:
@ale7714 your proposal makes sense - as long as the behavior is the same for users I think this sets us up for the future better and would allow us to return more information without breaking the contract again

EDIT: see below

cheukt:
@EvanDorsky brought up a good point. while no current clients are broken, we're still in a state where the behavior change means that clients have to infer current state instead of being able to use whether provisioning is up or down to figure out what actions can be taken.

will clients be able to take advantage of the smoother handling without further changes on agent? if no, I think it might make sense to make behavior here match main and then fast follow with the changes suggested.

PR #230, opened 2026-04-28
Bump tinygo.org/x/bluetooth v0.11.0 -> v0.15.0; use RemoveService

ale7714:
Replaces the 10000-iteration UnregisterApplication loop in removeServices with a single Adapter.RemoveService call.

The Service struct now carries an unexported id set by AddService and read back by RemoveService, so we keep a pointer to the service we registered (new Subsystem.bleService field) and hand it back on teardown.

org.bluez.Error.DoesNotExist (BlueZ already cleaned up) is treated as success, mirroring how Stop() handles "advertisement not started". The previous loop returned success only if at least one of 10000 unregister calls happened to succeed, so an already-clean adapter would surface an error; with the explicit handle that ambiguity goes away.

Cleanup error propagation in stopProvisioning() is unchanged from main and will be addressed in a follow-up PR.

Other relevant v0.15 behavior (verified via direct diff, not just changelog):
- adv.Configure now also writes org.bluez.Adapter1.Alias when LocalName is non-empty. The alias persists in BlueZ state but is only externally visible while Discoverable=true, which disablePairing flips off during stopProvisioningBluetooth -- so no leak in practice.
- adv.Stop still returns the unexported errAdvertisementNotStarted sentinel (not a wrapped dbus.Error) on org.bluez.Error.DoesNotExist, so the existing string compare remains the only viable check; left untouched.
- adv.Start spawns a dbus signal goroutine only when SetConnectHandler has been called -- the agent never sets one (pairing runs through bluetooth_tethering_linux.go via godbus directly), so this branch is dead for us.
- UUID changed from [4]uint32 to a struct and Bytes() became unexported; the agent only uses bluetooth.NewUUID and embeds UUIDs in struct literals, so no source change needed.

Drops the [APP-8081](https://viam.atlassian.net/browse/RSDK-10814) TODO (RemoveService is the upstream API the TODO was waiting for) and the now-unused getBluetoothDBus call from the removeServices path.


ale7714:
I think I'd like to land this PR before the BLE one, not a hard requirement.

subsystems/networking/bluetooth_linux.go
```diff
@@ -153,28 +153,24 @@ func getBluetoothDBus() (*dbus.Conn, dbus.BusObject, error) {
...
-
-	if ok {
-		return nil
+	dErr := &dbus.Error{}
+	if errors.As(err, dErr) && dErr.Name == "org.bluez.Error.DoesNotExist" {
+		// BlueZ already cleaned up the service (e.g. adapter power-cycled). Nothing to do.
+		return
 	}
-	return errors.New("could not find previous gatt service to remove")
+	n.logger.Errorf(
```
abe-winter: you're replacing the returned error with this log. is this because it's not recoverable?

how will the agent behave when bleService=nil but the old service has not been unregistered?

out-of-scope question: how do we want to communicate 'persistent error state' or 'needs restart for maintenance' in the UI?

subsystems/networking/bluetooth_linux.go
```diff
@@ -153,28 +153,24 @@ func getBluetoothDBus() (*dbus.Conn, dbus.BusObject, error) {
...
 
-func (n *Subsystem) removeServices() error {
-	_, adapter, err := getBluetoothDBus()
-	if err != nil {
-		return err
+func (n *Subsystem) removeServices() {
+	if n.bleService == nil {
+		return
 	}
+	defer func() { n.bleService = nil }()
```
abe-winter: do you still want to set bleService=nil in the error case (the non-DoesNotExist error case I mean)

abe-winter (approved):
lgtm if tested, left high level questions

cheukt:
@EvanDorsky fyi

cheukt (approved):
changes seem reasonable if tested

subsystems/networking/bluetooth_linux.go
ale7714: there's a real edge case here where we fail to `setProvisioning(false)` in the main loop on this error. but i'm gonna follow up on a separate pr. going to keeps this pr about bumping tinygo and using remove service

EvanDorsky:
I'll run the new BLE serial test suite against this but looks good.

ale7714:
i think this is pretty safe and i'm going to merge so i can rebase the other PR

PR #234, opened 2026-05-05
APP-16234: add offline msi installer with Agent binary embedded

aldenh-viam:
1) Download `viam-agent-stable-windows-x86_64.exe` to `msi\viam-agent-from-installer.exe`
2) Run `make msi`
3) Distribute generated `msi\bin\x64\Release\Package.msi`

(above instructions should be set up in CI)

I confirmed I'm able to install Agent on an offline machine using the installer. It writes the Viam directories, including:
```
C:\opt\bin\viam-agent (symlink)
C:\opt\cache\viam-agent-from-installer.exe
```
and starts up viam-agent.

`viam-agent-stable-windows-x86_64.exe` is 28,194KB and `Package.msi` is 9,060 KB:
> MSI cab files are compressed. <MediaTemplate EmbedCab="yes" /> (line 19 of Package.wxs) embeds a CAB inside the MSI, and by default WiX uses LZX/MSZIP compression. Go binaries compress well (typically 3–4× ratio) because they contain lots of repeated runtime metadata, symbol tables, and zero-padded sections. 28MB → 9MB ≈ 32% is normal.


cheukt:
we should build and upload the msi to gcp - do you want to do it in this pr or as a fast follow?

aldenh-viam:
> we should build and upload the msi to gcp - do you want to do it in this pr or as a fast follow?

How are we currently building and distributing the (online) `msi` installer? I don't see anything about it in the `.github/workflows` dir.

cheukt:
@abe-winter would know more, but i suspect we don't have a process for it yet

abe-winter:
yup no yaml yet for msi. remaining work to normalize on msi:
- test that firewalling is correct in tcp mode 
- add msi yaml
- update docs 
- remove old exe build at some point

EvanDorsky:
Worked great on my windows vm, way better install experience than before.

aldenh-viam:
Given the compressed offline installer is only 9MB, should I just make these changes in `msi/` so we don't need to maintain both online and offline msi installers?

cheukt:
ya lets do that

cheukt:
where are we pushing the msi?

.github/workflows/build-agent-msi.yaml
```diff
@@ -0,0 +1,32 @@
...
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Download agent binary
+        shell: pwsh
+        run: |
+          $ProgressPreference = 'SilentlyContinue'
+          Invoke-WebRequest `
+            -Uri "https://storage.googleapis.com/packages.viam.com/apps/viam-agent/viam-agent-stable-windows-x86_64" `
```
abe-winter: 1. this binary comes from workflow.yml in this repo -- in theory you can build them at the same time rather than downloading
2. **warning**: if you run this download-based msi task as part of an agent release workflow, I think you will get the previous stable version for weird release flow reasons
3. curl is available on this machine if you prefer

abe-winter (approved):

.github/workflows/build-agent-msi.yaml
cheukt: ideally we should use the binary that we just built

aldenh-viam:
CI run sample: https://github.com/viamrobotics/agent/actions/runs/25498673196/job/74826372596

It would upload `viam-agent-v0.27.3-dev.9-windows-x86_64.msi` to GCS.

.github/workflows/workflow.yaml
```diff
@@ -137,6 +145,44 @@ jobs:
...
+        name: viam-agent-msi
+        path: viam-agent-*-windows-x86_64.msi
+    - uses: google-github-actions/auth@v2
+      with:
+        credentials_json: ${{ secrets.GCP_CREDENTIALS }}
+    - uses: google-github-actions/setup-gcloud@v2
+    - name: Upload MSI to GCS (Release)
+      if: github.event_name == 'release'
+      shell: bash
+      run: gsutil -h "Cache-Control:no-cache" cp viam-agent-*-windows-x86_64.msi gs://packages.viam.com/apps/viam-agent/
```
cheukt: can we also upload to viam-agent-stable-windows-x86_64.msi for stable releases? so we have a stable path to refer to as well

cheukt (commented):
looks good! just one more request

cheukt (approved):

PR #235, opened 2026-05-06
RSDK-13797: Add managed os updates support on rpm and apt based Linux systems

jmatth:
- Adds two new fields to the existing config field for auto os updates: `managed-all` and `managed-security`
- Updates the existing unattended upgrades functionality on Debian to disable itself when one of the new options is specified
- Adds logic for to manage os updates on apt and rpm based systems by
  - Installing updates on an interval configurable in hours
  - Rebooting the machine within the next maintenance window if a package update indicates a reboot is needed
- Makes incremental changes to remove some stringly typed code that was bothering me


subsystems/syscfg/managed_upgrades_apt_linux.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+
+	if securityOnly {
+		return a.runSecurityUpgrade(ctx)
+	}
+	return a.runFullUpgrade(ctx)
+}
+
+func (a aptPackageManager) runFullUpgrade(ctx context.Context) error {
+	return pkgCmd(ctx, "apt-get", "upgrade", "-y",
```
aldenh-viam: Is `dist-upgrade` out of scope?

subsystems/syscfg/managed_upgrades_apt_linux.go
```diff
@@ -0,0 +1,86 @@
...
+	if _, err := utils.WriteFileIfNew(autoUpgradesPath, []byte(autoUpgradesContentsEnabled)); err != nil {
+		return errw.Wrap(err, "writing auto-upgrades config")
+	}
+
+	cmd := exec.CommandContext(ctx, "unattended-upgrade", "--verbose")
+	cmd.Env = append(os.Environ(),
+		"DEBIAN_FRONTEND=noninteractive",
+		"APT_LISTCHANGES_FRONTEND=none",
+	)
+	output, err := cmd.CombinedOutput()
```
aldenh-viam, line 81: Sorry to be that guy, but would it be possible for this to hang and leave the system in an unknown state with no logs, e.g. if your network drops midway through?

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/viamrobotics/agent
 
-go 1.25.1
+go 1.26
```
aldenh-viam, line 3: Might be worth quickly investigating that this doesn't result in any strange behaviors. I was playing with it recently on two separate occasions and found that 1.26 caused some RDK tests to fail, and also the new GC was slower and not more CPU efficient in a module where I would expect it to help.

aldenh-viam (commented):
First pass LGTM so far, just some thoughts. I'll take a another closer look once it's ready for review.

subsystems/syscfg/managed_upgrades_apt_linux.go
jmatth: As far as I'm aware apt and rpm first download all requrired packages and then install them in separate steps, so a network outage while downloading should just make the update fail before any changes are made to the system. There is always the possibility the machine loses power or the process gets SIGKILLed in the middle of the actual install though, which isn't something we can really prevent. When that happens on apt based systems, future invocations of apt-get detect the incomplete transaction and print an error that includes a dpkg command that can be run to complete any pending operations and return the system to a clean state. I'm not sure if rpm has similar incomplete transaction recovery. I guess it's also always possible that a failure occurs while something critical like the kernel is being written to disk, but if that happens I really don't think there's anything we can do.

subsystems/syscfg/managed_upgrades_apt_linux.go
jmatth: Yes, the scope doc calls out upgrades between major versions as out of scope

subsystems/syscfg/managed_upgrades_linux.go
```diff
@@ -0,0 +1,203 @@
...
+
+	if cancel != nil {
+		cancel()
+		s.upgradeWorker.Wait()
+	}
+}
+
+// NeedsOSReboot returns true if a system reboot is pending due to installed
+// package updates.
+func (s *Subsystem) NeedsOSReboot(ctx context.Context) bool {
```
jmatth, line 81: In the original claude generated code this just returned `s.needsOSReboot`, which in turn was set after an upgrade was performed. I changed it to this caching flow where we perform the full check + cache the first positive result to guard against issues like viam-agent crashing or this goroutine losing a race with a self-update restart and getting stuck in a bad state where a reboot is requried but the current agent process can't tell.

subsystems/syscfg/upgrades_linux.go
```diff
@@ -79,31 +94,35 @@ func (s *Subsystem) EnforceUpgrades(ctx context.Context) error {
...
 	}
 	return nil
 }
 
-func checkSupportedDistro() error {
+// checkSupportedDistro checks if the current system supports unattended
+// upgrades. It returns a empty string on supported systems, and a string
+// explaining why there is no support otherwise. If it is unable to gather the
+// required information to check for support an error is returned.
+func checkSupportedDistro() (string, error) {
```
jmatth, line 108: I don't particularly love this function contract but using a single error consistently produced misleading logs on rpm based systems.

jmatth (commented):
@aldenh-viam ok, I think this is ready for review now. Testing so far is unfortunately all manual but I verified that `managed-all` and `managed-security` work on both Debian 13 and Alma 10 VMs. The reboot logic respects maintenance windows and correctly detects when a reboot is required even if viam-agent has been restarted since applying the update or the update was performed manually outside of the agent process.

go.mod
jmatth: Degraded performance with the new GC is unfortunate since it's set to become the only GC in 1.27. All the tests pass with the new version here but I don't know how much that counts for in this repo. I'll try bumping rdk to 1.26 to see what kinds of issues it causes and if any might be relevant here

aldenh-viam (approved):

PR #236, opened 2026-05-12
Add viam-server manager tests

cheukt:
Slight refactoring to inject a fake, but also streamline the manager code so that it's slightly easier to reason about


danielbotros (approved):
LGTM!

PR #237, opened 2026-05-13
Fallback if systemd-path or systemd-search-system-unit don't exist

danielbotros:
@johnwalicki and I were helping a customer (Miele) install agent on a Linux machine running Ubuntu 20.04. The script exited fatally with this error message:  `running 'systemd-path systemd-search-system-unit' output: Path systemd-search-system-unit not known`:
image

It seems like they had an outdated version of `systemd` which doesn't recognize `systemd-search-system-unit` and we don't gracefully handle this scenario. We already have a fallback path (`/etc/systemd/system`) when we cannot find the default service path (`/usr/local/lib/systemd/system`), so this PR just makes it so we use that fallback path when `systemd-path` or `systemd-path systemd-search-system-unit` errors out instead of fatally exiting.

### DIFF
--- a/utils/systemd/manager.go
+++ b/utils/systemd/manager.go
@@ -117,10 +117,7 @@ func (s *SystemdManager) getServiceFilePath(ctx context.Context, serviceFile str
 	// Check if the new, preferred path is present in the systemd search path and
 	// therefore a migration may need to be performed.
 	searchPaths, err := s.SystemdSearchPaths(ctx)
-	if err != nil {
-		return "", false, err
-	}
-	if !slices.Contains(searchPaths, s.dirs.serviceFileDir) {
+	if err != nil || !slices.Contains(searchPaths, s.dirs.serviceFileDir) {
 		s.logger.Warnf(
 			"Systemd does not have %s in its unit search path, installing directly to %s",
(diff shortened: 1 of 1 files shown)


johnwalicki (approved):
Especially useful when installing on Ubuntu 20.04 and old systemd
Thanks for the fix

utils/systemd/manager.go
```diff
@@ -118,7 +118,11 @@ func (s *SystemdManager) getServiceFilePath(ctx context.Context, serviceFile str
 	// therefore a migration may need to be performed.
 	searchPaths, err := s.SystemdSearchPaths(ctx)
 	if err != nil {
-		return "", false, err
+		s.logger.Warnf(
+			"Could not determine systemd unit search paths (%v); installing service file directly to %s",
+			err, s.dirs.fallbackFileDir,
+		)
```
jmatth: Prefer structured logs in new code:


```suggestion
		s.logger.Warnw(
			"Could not determine systemd unit search paths; installing to fallback path",
			"err", err,
			"installPath", oldFilePath,
		)

```

jmatth (approved):

PR #239, opened 2026-05-15
RSDK-13711 Specify test version

EvanDorsky:
This PR re-engineers the test suite to enable testing against a specified version of `viam-agent` instead of `stable`.

## Interface/infra changes
- new config value: `viam_agent_test` to specify the version under test
- `installAgent` now uses the local `install.sh` (just like `uninstallAgent`) instead of pulling the install script from gcp.
- `installAgent` now checks that viam-agent is running at the specified version, instead of just running at all. If the version isn't correct, it installs the specified version.
- There is a new way to specify the version in test cases, called "the version under test"
- The after suite hook now pins agent back to the version under test, instead of stable
- `serialclient.EnsureOnline` was re-added to the pre-scenario hook (this fixes some instability caused by the provisioning tests leaving the network adapter in a weird state)
- new code to retrieve specific versions based on the high level version spec, and gcs url based on specific versions (detailed below)

The agent version under test can be specified in a number of ways:
- `pr.xxx` will test a PR by number, if it has the `dev-release` label (so that artifacts are built) e.g. `"pr.227"`
- `"dev"` will test the tip of main
- `"stable"` will test the latest stable release
- `"x.y.z"` will test any release version e.g. `"0.27.3"`
- `"file:///path/to/agent"` will test an arbitrary binary (that you manually copy onto the pi)

Each feature test now prints the versions under test before running:
```
=================================================
=================================================
    Testing Agent Version: dev (0.27.3-dev.8)

          Stable Agent Version: 0.27.3
         Stable Server Version: 0.123.0
=================================================
=================================================
```

## Test changes

Tests have been modified to account for the new syntax requirements to specify versions (like `stable`, `the version under test`, `an old version`):
- `viam-agent is installed` -> `viam-agent is installed at the version under test`
- `the viam-agent systemd unit is running` -> `the viam-agent systemd unit is running with the version under test`

The upgrade and downgrade tests have been changed so they make sense with this new flow:
- `upgrade.feature` upgrades from stable to the version under test
- `downgrade.feature` downgrades from the version under test to stable

## Changes beyond tests
This PR makes a small modification to `install.sh` - it adds `$AGENT_CUSTOM_URL`, which will be used instead of `$URL` as the download target for the agent binary if it `$AGENT_CUSTOM_URL` is passed in. This should be totally transparent if not used, but it enables the test to install any version (instead of relying on agent to update itself via pinning, which isn't really the same thing).


agent_serial_test.go
```diff
@@ -185,24 +196,62 @@ func InitializeSuite(t *testing.T) func(*godog.TestSuiteContext) {
...
 	// Restart viam-agent before each scenario (if it is running) so that every
 	// scenario starts with a fresh systemd InvocationID. This ensures that
 	// journal-based checks cannot match log lines produced by a previous scenario.
 	ctx.Before(func(ctx context.Context, _ *godog.Scenario) (context.Context, error) {
+		if cfg.Versions.Test != "" {
+			concrete, err := resolveVersionSpec(ctx, cfg.Versions.Test)
+			concreteTestVersion = concrete
+			logger.Infof("Version under test: %s\n", concreteTestVersion)
+			if err != nil {
+				panic(fmt.Errorf("resolving install version %q: %w", cfg.Versions.Test, err))
```
danielbotros: nit: handle error before logging return value

agent_serial_test.go
```diff
@@ -185,24 +196,62 @@ func InitializeSuite(t *testing.T) func(*godog.TestSuiteContext) {
...
+		consoleWidth := len(slices.MaxFunc(testMsgs, func(a, b string) int { return len(a) - len(b) })) + 8
+		fmt.Println(strings.Repeat("=", consoleWidth))
+		fmt.Println(strings.Repeat("=", consoleWidth))
+		centerPrint(testMsgs[0], consoleWidth)
+		fmt.Println()
+		for _, m := range testMsgs[1:] {
+			centerPrint(m, consoleWidth)
+		}
+		fmt.Println(strings.Repeat("=", consoleWidth))
+		fmt.Println(strings.Repeat("=", consoleWidth))
```
danielbotros: Should we move this outside the `ctx.Before` so it doesn't run before every scenario? Would save on time / unnecessary network traffic / spammy logs

agent_serial_test.go
```diff
@@ -808,24 +1029,59 @@ func bleSurfacesExpectedError(expectedErr string) error {
...
+		return ctx, err
+	}
+
+	logger.Infof("Install version: %s\n", concrete)
+	// Don't use the concrete test version if it's a file pin, because gcsURL can't handle file pins
+	if !strings.HasPrefix(concrete, "file://") {
+		cmd += fmt.Sprintf(" AGENT_CUSTOM_URL=%s", gcsURL("viam-agent", concrete))
+	}
+	cmd += " sh"
+	err = serialClient.RunScript(installScript, cmd).Error()
```
danielbotros, line 1078: This error gets lost / thrown out if we enter the `strings.HasPrefix(concrete, "file://")` branch below

features/serial/agent-reject-viam-server-binary.feature
```diff
@@ -1,7 +1,7 @@
 Feature: Pinning viam-agent to a viam-server binary is rejected
 
   Background:
-    Given viam-agent is installed
+    Given viam-agent is installed at the version under test
     And viam-agent is pinned to stable
     And the viam-agent systemd unit is running with stable
```
danielbotros: I think these need to be updated with `the version under test` too

agent_serial_test.go
```diff
@@ -845,15 +1101,16 @@ func testAgentNotFound(ctx context.Context) (context.Context, error) {
...
 	if err != nil {
 		return ctx, err
 	}
 	return testAgentState(ctx, "SubState", "running")
 }
 
-func testSystemdAgentStartVersion(ctx context.Context, version string) (context.Context, error) {
+func testSystemdAgentStartVersion(ctx context.Context, version string, wait bool) (context.Context, error) {
 	versionTest := versionStrToMatcher(version)
+	logger.Infof("Check for version %s with matcher %s\n", version, versionTest)
```
danielbotros: This tries to log `versionTest` (a function) as a string here.

agent_serial_test.go
```diff
@@ -862,13 +1119,19 @@ func testSystemdAgentStartVersion(ctx context.Context, version string) (context.
...
-			continue
+			if wait {
+				err = errors.New(check)
+				continue
+			} else {
+				return ctx, errors.New(check)
+			}
 		}
 		return ctx, nil
 	}
```
danielbotros: I think splitting this up into two functions (`checkAgentVersion` and `waitForAgentVersion` maybe?) would be cleaner and easier to reason about than the `wait` bool param gating diverging behavior

danielbotros (commented):
Much needed improvement, thanks Evan! Have not run tests yet but left some initial comments

agent_serial_test.go
EvanDorsky: I think `EnsureOnline` needs to be here because individual scenarios can mess up the adapter state. But yes it would be better if versions were printed only once at the beginning.

features/serial/agent-reject-viam-server-binary.feature
EvanDorsky: Good catch! You're right, I missed this one.

agent_serial_test.go
EvanDorsky: Yeah you're right that would be better.

install.sh
```diff
@@ -17,6 +17,9 @@ if [ "$OS" = "Darwin" ]; then
 	BINARY_OS_PREFIX="darwin-"
 fi
 URL="https://storage.googleapis.com/packages.viam.com/apps/viam-agent/viam-agent-stable-${BINARY_OS_PREFIX}${ARCH}"
```
jmatth, line 19: If you want to be cool and don't care about readability you can do this conditional assignment in one line:

```suggestion
URL="${AGENT_CUSTOM_URL}:-https://storage.googleapis.com/packages.viam.com/apps/viam-agent/viam-agent-stable-${BINARY_OS_PREFIX}${ARCH}}"
```

agent_serial_test.go
```diff
@@ -286,15 +340,161 @@ func hostEnsureOnline(ctx context.Context) (context.Context, error) {
...
+	}
+	sort.Slice(body.Items, func(i, j int) bool {
+		return body.Items[i].TimeCreated.After(body.Items[j].TimeCreated)
+	})
+	return body.Items, nil
+}
+
+// latestStableRelease returns the most recently uploaded stable viam-agent
+// release version (bare semver, e.g. "0.27.3").
+func latestStableRelease(ctx context.Context) (string, error) {
```
jmatth, line 446: It might be possible to remove the need to specify the stable + dev versions in the config and just replace them whith these helpers.

agent_serial_test.go
```diff
@@ -286,15 +340,161 @@ func hostEnsureOnline(ctx context.Context) (context.Context, error) {
...
+// resolveVersionSpec turns a TOML version specifier into a concrete version
+// string. Recognized forms:
+//   - "stable"  -> latest stable release       (e.g. "0.27.3")
+//   - "dev"     -> latest main-branch dev build (e.g. "0.27.3-dev.5")
+//   - "pr.<N>"  -> latest dev-release for PR N  (e.g. "0.27.3-pr.227.<sha>")
+//   - anything else is assumed to already be concrete and returned as-is.
+//
+// Only viam-agent specifiers are supported today; viam-server has different
+// upload paths and would need its own resolver.
+func resolveVersionSpec(ctx context.Context, spec string) (string, error) {
```
jmatth, line 383: This gets called by `installAgent`, which in turn is called multiple times across the test suite. It would be nice to cache the resolved version to cut down on unnecessary GCS queries but not essential.

jmatth (approved):
LGTM. The few comments I left are nitpicks and don't need to be addressed to merge.

danielbotros (approved):
LGTM, congrats on finishing agent testing 😄!!

PR #244, opened 2026-05-26
APP-16486 - viam server running version in app's agent_subsystem_versions should be accurate

cheukt:
technically the ViamServerRunningVersion is a reflection of the desired running version (since it will be populated even if the viam-server has not successfully started up), but this change is better than what it is now.

maybe we should push the viam-server state too (running, initializing, etc)

### DIFF
--- a/manager.go
+++ b/manager.go
@@ -307,6 +307,7 @@ func (m *Manager) SubsystemUpdates(ctx context.Context) {
 					m.logger.Warn(err)
 				} else {
 					m.viamServerNeedsRestart = false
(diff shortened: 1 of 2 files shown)


benjirewis (approved):
LGTM.

More broadly though: what's the point of `viam-agent` and `viam-server` both submitting a `viam-server` version? Is another possible solution to _stop_ reporting `viam-server` versions from `viam-agent`? Especially if the values aren't used anywhere in the frontend and the `viam-server` versions reported by agent aren't as accurate as the one reported by `viam-server` itself.

PR #245, opened 2026-05-28
Mark viam-server running version on startup

cheukt:
I couldn't make a dev release for my last pr because I pushed to my fork instead of this, so this pr fixes an issue that I found very quickly after testing

the running version wouldn't be marked until the first time viam-server stopped

### DIFF
--- a/manager.go
+++ b/manager.go
@@ -307,7 +307,6 @@ func (m *Manager) SubsystemUpdates(ctx context.Context) {
 					m.logger.Warn(err)
 				} else {
 					m.viamServerNeedsRestart = false
-					m.cache.MarkViamServerRunningVersion()
 				}
 				if m.viamAgentNeedsRestart {
 					m.Exit(fmt.Sprintf("A new version of %s has been installed", SubsystemName))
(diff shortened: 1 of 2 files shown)


manager.go
```diff
@@ -317,6 +316,9 @@ func (m *Manager) SubsystemUpdates(ctx context.Context) {
 				m.logger.Warnf("%s has NOT allowed a restart; will NOT restart", viamserver.SubsysName)
 			}
 		}
+		if !m.viamServerNeedsRestart {
+			m.cache.MarkViamServerRunningVersion()
```
benjirewis, line 320: [q] How is this different from what we initially had (I mean a week ago before your last PR/commit)? I'm having trouble parsing the `if m.viamServerNeedsRestart || m.viamAgentNeedsRestart` followed by `if !m.viamServerNeedsRestart {`. With which case are we _not_ marking a running `viam-server` version with your changes that we _were_ previously marking?

cheukt: what we initially had was that we were always marking a running a `viam-server` version, which is bad if the `viam-server` didn't actually get stopped/get restarted.

now, if stopping viam-server failed, m.viamServerNeedsRestart will stay true and we won't be marking the running viam-server version.

that's the intended change anyway

benjirewis (approved):

manager.go
benjirewis: > if stopping viam-server failed, m.viamServerNeedsRestart will stay true and we won't be marking the running viam-server version.

Got it; that makes sense to me. Thanks!

PR #249, opened 2026-06-09
Add Windows managed upgrades support

jmatth:
- Adds a `syscfg` implementation on Windows that only implements the new managed upgrades support
- Installing Windows updates is managed via a Powershell module that must be pulled from the internet if it is not already installed
- Changes how the `syscfg` and `networking` subsystems are enabled so we don't have to modify the Windows service on existing installs. I think this is the ugliest part of the change

Tested manually on Windows 11

### DIFF
--- a/cmd/viam-agent/main.go
+++ b/cmd/viam-agent/main.go
@@ -46,7 +46,7 @@ type agentOpts struct {
 	Debug                     bool   `description:"Enable debug logging (agent only)"                                                            env:"VIAM_AGENT_DEBUG"                           long:"debug"    short:"d"`
 	ViamDir                   string `description:"Use a custom path for agent directories"                                                      long:"viam-dir"`
 	EnableSyscfgSubsystem     *bool  `description:"Enable system configuration management subsystem"                                             long:"enable-syscfg"`
-	EnableNetworkingSubsystem bool   `description:"Enable networking management subsystem"                                                       long:"enable-networking"`
+	EnableNetworkingSubsystem *bool  `description:"Enable networking management subsystem"                                                       long:"enable-networking"`
 	UpdateFirst               bool   `description:"Update versions before starting"                                                              env:"VIAM_AGENT_WAIT_FOR_UPDATE"                 long:"wait"     short:"w"`
 	DevMode                   bool   `description:"Nothing (deprecated and will be removed in a future release)"                                 long:"dev-mode"`
 	Help                      bool   `description:"Show this help message"                                                                       long:"help"                                      short:"h"`
@@ -132,7 +132,10 @@ func commonMain(runningAsService bool) {
 		utils.CLIEnableSyscfgSubsystem = true
 	}
 
-	if opts.EnableNetworkingSubsystem {
+	// Similar to above, prefer the more specific `enable-networking` flag if it
+	// is set, otherwise use `as-service`
+	enableNetworkingSubsystemFlag := mo.PointerToOption(opts.EnableNetworkingSubsystem)
+	if enableNetworkingSubsystemFlag.OrEmpty() || (enableNetworkingSubsystemFlag.IsAbsent() && asServiceFlag.OrEmpty()) {
 		utils.CLIEnableNetworkingSubsystem = true
 	}
 
--- a/cmd/viam-agent/main.go
+++ b/cmd/viam-agent/main.go
@@ -134,12 +134,11 @@ func commonMain(runningAsService bool) {
 		utils.CLIEnableSyscfgSubsystem = true
 	}
 
-	// Similar to above, prefer the more specific `enable-networking` flag if it
-	// is set, otherwise use `as-service`. At time of writing this should never
-	// happen because `runningAsService` is only set on Windows we don't support
-	// networking on Windows, but the code is structured this way for consistency
-	// in case other platforms pass the argument on we add Windows networking
-	// support in the future.
+	// Same as above, but for `--enable-networking`. At time of writing this
+	// should never happen because `runningAsService` is only set on Windows and
+	// we don't support networking on Windows, but the code is structured this
+	// way for consistency in case other platforms pass the argument or we add
+	// Windows networking support in the future.
 	enableNetworkingSubsystemFlag := mo.PointerToOption(opts.EnableNetworkingSubsystem)
 	if enableNetworkingSubsystemFlag.OrEmpty() || (enableNetworkingSubsystemFlag.IsAbsent() && runningAsService) {
 		utils.CLIEnableNetworkingSubsystem = true
--- a/cmd/viam-agent/main.go
+++ b/cmd/viam-agent/main.go
@@ -53,7 +53,7 @@ type agentOpts struct {
 }
 
 //nolint:gocognit
-func commonMain() {
+func commonMain(runningAsService bool) {
 	// enable debug logging in case it is needed before args are parsed successfully.
 	if os.Getenv("VIAM_AGENT_DEBUG") != "" {
 		globalLogger.SetLevel(logging.DEBUG)
@@ -118,7 +118,7 @@ func commonMain() {
 		utils.CLIWaitForUpdateCheck = true
 	}
 
-	if opts.EnableSyscfgSubsystem {
+	if opts.EnableSyscfgSubsystem || runningAsService {
 		utils.CLIEnableSyscfgSubsystem = true
 	}
 
--- a/cmd/viam-agent/main.go
+++ b/cmd/viam-agent/main.go
@@ -40,18 +40,18 @@ var (
 
 //nolint:lll
 type agentOpts struct {
-	Config                    string `default:"/etc/viam.json"                                                   description:"Path to machine credentials file"   long:"config"   short:"c"`
-	DefaultsConfig            string `default:"/etc/viam-defaults.json"                                          description:"Path to manufacturer defaults file" long:"defaults"`
-	Install                   bool   `description:"Install systemd service"                                      long:"install"`
-	Debug                     bool   `description:"Enable debug logging (agent only)"                            env:"VIAM_AGENT_DEBUG"                           long:"debug"    short:"d"`
-	ViamDir                   string `description:"Use a custom path for agent directories"                      long:"viam-dir"`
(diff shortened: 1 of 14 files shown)


cmd/viam-agent/main.go
```diff
@@ -118,15 +120,21 @@ func commonMain() {
...
 	}
 
-	if opts.EnableSyscfgSubsystem {
+	// Cases are as follows:
+	// 1. `--enable-syscfg` passed on command line. Used by developers specifically testing the syscfg subsystem.
+	// 2. `--as-service` flag passed on command line. Used by systemd unit.
+	// 3. `runningAsService` function argument is true. Used on Windows machines.
+	asServiceFlag := mo.PointerToOption(opts.AsService)
+	enableSyscfgSubsystemFlag := mo.PointerToOption(opts.EnableSyscfgSubsystem)
+	if enableSyscfgSubsystemFlag.OrEmpty() || (enableSyscfgSubsystemFlag.IsAbsent() && asServiceFlag.OrEmpty()) || (enableSyscfgSubsystemFlag.IsAbsent() && asServiceFlag.IsAbsent() && runningAsService) {
```
jmatth: I don't love this but right now the Windows service is set up by the Windows installer and not modified during updates. I updated the Windows `main` to pass a bool indicating whether its running as a windows process but then that led down a rabbit hole of "but what if we're running as a service but in a non-standard install that wants to disable `syscfg`?".

The `--as-service` flag is to guard against having to update the service too frequently in the future. If we add more subsystems that should be off during development but enabled by default on real machines we can group them under this flag rather than updating all the service definitions.

EvanDorsky: I think it's fair enough to use a flag like this, but looks like there's a bug right now where `--as-service` isn't wired to enable the networking subsystem the way it's wired into syscfg here. `--as-service` should be equivalent to `--enable-syscfg --enable-networking` on Linux, right?

subsystems/syscfg/managed_upgrades_windows.go
```diff
@@ -0,0 +1,167 @@
...
+		// reboot happens.
+		return true
+	}
+
+	if !isManaged(autoUpgradeType) {
+		// We only care about managing reboots in managed upgrade mode.
+		return false
+	}
+
+	needsReboot := windowsRebootRequired(ctx)
```
EvanDorsky: Is this intended to be a different variable from `needReboot`?

jmatth: Nope. I think autocomplete was a bit too helpful there :sweat_smile:

cmd/viam-agent/main.go
jmatth: Correct. I think I `git reset` that while waffling on how to structure this code and forgot to add it back in. Fixing now

subsystems/syscfg/managed_upgrades_windows.go
```diff
@@ -0,0 +1,167 @@
...
+		s.runManagedUpgrade(upgradeCtx)
+
+		ticker := time.NewTicker(interval)
+		defer ticker.Stop()
+		for {
+			select {
+			case <-upgradeCtx.Done():
+				return
+			case <-ticker.C:
+				s.runManagedUpgrade(upgradeCtx)
```
aldenh-viam, line 77: Would it be bad if this takes longer than than the interval?

subsystems/syscfg/managed_upgrades_windows.go
```diff
@@ -0,0 +1,167 @@
...
+// startManagedUpgrades launches the background goroutine that periodically runs Windows Update.
+// Must be called while s.mu is held.
+func (s *Subsystem) startManagedUpgrades(ctx context.Context) {
+	if s.upgradeCancel != nil {
+		return // already running
+	}
+
+	interval := time.Duration(float64(time.Hour) * s.cfg.OSManagedUpgradeIntervalHours)
+	if interval < time.Hour {
+		interval = defaultUpgradeInterval
```
aldenh-viam: I guess this is to set the default interval if `OSManagedUpgradeIntervalHours` isn't specified, but if it is something like 0.5, shouldn't we set it to `time.Hour` (if that's the minimum) rather than revert to the default?

subsystems/syscfg/managed_upgrades_linux.go
```diff
@@ -6,16 +6,15 @@ package syscfg
 // and coordinate reboots with viam-server's maintenance window.
 //
 // Supported package managers (tried in preference order):
-//   - dnf  – Fedora, RHEL 8+, Rocky Linux, AlmaLinux, etc.
-//   - apt-get – Debian, Ubuntu, Raspberry Pi OS, etc.
-//   - yum  – RHEL 7, CentOS 7
+//   - dnf  - Fedora, RHEL 8+, Rocky Linux, AlmaLinux, etc.
+//   - apt-get - Debian, Ubuntu, Raspberry Pi OS, etc.
+//   - yum  - RHEL 7, CentOS 7
```
aldenh-viam, line 11: :godmode:

cmd/viam-agent/main.go
```diff
@@ -39,21 +40,22 @@ var (
...
+	UpdateFirst               bool   `description:"Update versions before starting"                                                              env:"VIAM_AGENT_WAIT_FOR_UPDATE"                 long:"wait"     short:"w"`
+	DevMode                   bool   `description:"Nothing (deprecated and will be removed in a future release)"                                 long:"dev-mode"`
+	Help                      bool   `description:"Show this help message"                                                                       long:"help"                                      short:"h"`
+	Version                   bool   `description:"Show version"                                                                                 long:"version"                                   short:"v"`
+	AsService                 *bool  `description:"Indicate viam-agent is running as a process. Implies --enable-syscfg and --enable-networking" long:"as-service"`
 }
 
 //nolint:gocognit
-func commonMain() {
+func commonMain(runningAsService bool) {
```
aldenh-viam: Would it be simpler if only Windows uses AsService (programmatically set if I'm understanding correctly) and others continue using the `--enable-syscfg/networking` cli flags?

subsystems/syscfg/syscfg_windows.go
```diff
@@ -0,0 +1,101 @@
...
+
+	if s.started {
+		return nil
+	}
+	s.logger.Info("Starting syscfg subsystem")
+	s.started = true
+	s.healthy = true
+
+	mode := s.cfg.OSAutoUpgradeType
+	if mode == managedSecurityMode || mode == "managed-all" {
```
aldenh-viam: There are some constants for these strings in `utils/config.go`

aldenh-viam (approved):

subsystems/syscfg/managed_upgrades_windows.go
jmatth: It should be fine. This call blocks so we won't try to run another upgrade in parallel. Worst case is probably it completes right before the end of the interval and ends up running another upgrade check immediately, which might be a bit annoying but wouldn't cause any real issues

subsystems/syscfg/managed_upgrades_linux.go
jmatth: [I have ~special eyes~ a plugin][1]

[1]: https://marketplace.visualstudio.com/items?itemName=nhoizey.gremlins

EvanDorsky (approved):
Lgtm in terms of implementation and with the cautious rollout plan that has been laid out.

I'm most concerned about machines with poor internet connections, limited disk space, and potentially very out of date windows installs (which could trigger large downloads). But as long as we can manage the deploy to those machines, good otherwise.

subsystems/syscfg/syscfg_windows.go
jmatth: Not only are there, I already used them in some helper methods on the Linux side that could be reused here :facepalm:. I'm gonna take another pass at deduping the Linux and Windows implementation, there's plenty of shared code but just enough differences to make deduplicating the code annoying

cmd/viam-agent/main.go
jmatth: Probably. I was trying to keep the behavior consistent across platforms and guard against another situation where we want to change the default behavior but doing so on Windows is annoying because the installer manages the service configuration. I may have gotten carried away though, let me see about switching to just using `runningAsService` in a way that doesn't ignore an explicit `--enable-syscfg=false` flag

cmd/viam-agent/main.go
```diff
@@ -122,20 +128,20 @@ func commonMain(runningAsService bool) {
...
 	// 1. `--enable-syscfg` passed on command line. Used by developers specifically testing the syscfg subsystem.
-	// 2. `--as-service` flag passed on command line. Used by systemd unit.
-	// 3. `runningAsService` function argument is true. Used on Windows machines.
-	asServiceFlag := mo.PointerToOption(opts.AsService)
+	// 2. `runningAsService` function argument is true. Used on Windows machines.
 	enableSyscfgSubsystemFlag := mo.PointerToOption(opts.EnableSyscfgSubsystem)
-	if enableSyscfgSubsystemFlag.OrEmpty() ||
-		(enableSyscfgSubsystemFlag.IsAbsent() && asServiceFlag.OrEmpty()) ||
-		(enableSyscfgSubsystemFlag.IsAbsent() && asServiceFlag.IsAbsent() && runningAsService) {
+	if enableSyscfgSubsystemFlag.OrEmpty() || (enableSyscfgSubsystemFlag.IsAbsent() && runningAsService) {
```
aldenh-viam, line 133: Is this the same as 
```go
if enableSyscfgSubsystemFlag.OrEmpty() || runningAsService) {
```
?

aldenh-viam (approved):

cmd/viam-agent/main.go
jmatth: No. `enableSyscfgSubsystemFlag.OrEmpty() || runningAsService` will evaluate to true even if agent is run as a windows service with `--enable-syscfg=false` passed to the command. `enableSyscfgSubsystemFlag.OrEmpty() || (enableSyscfgSubsystemFlag.IsAbsent() && runningAsService)` lets `--enable-syscfg` take priority if it is set, and only falls back on `runningAsService` if `--enable-syscfg` not provided. I'll admit that I don't think that use case is likely to come up, but I didn't want to leave the code in a state where that flag is ignored and it's impossible to disable the subsystem when running as a windows service

PR #252, opened 2026-06-10
Do not let mise set GOROOT in CI

jmatth:
This fixes an issue where the go binary installed with `setup-go` would run but with a `GOROOT` pointing to the version installed by mise. This broke the build when they chose different versions. I would like to remove `setup-go` but the mise action's caching behavior currently doesn't allow for updating to the latest version of a tool that still satisfies the project's version constaints on a cache hit


ale7714 (approved):
thank you!

PR #256, opened 2026-06-16
Skip restart check and config fetch if connection state is TRANSIENT_FAILURE

aldenh-viam:
Greatly reduces amount of these logs in the local journalctl when offline:

```
(every 1s)   WARN    viam-agent    fetching viam-agent config: rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing: dial tcp: lookup app.viam.com ...
(every 5s)   WARN    viam-agent    checking if restart needed: rpc error: code = Unavailable desc = connection error: desc = "transport: Error while dialing: dial tcp: lookup app.viam.com ...
```

Note however, if we start while completely offline, it fails even earlier at the `dial`, so that will spam. I will take a closer look at it.


go.mod
```diff
@@ -41,9 +41,19 @@ require (
 require (
```
cheukt, line 41: do you know where all the go mod changes are coming from?

manager.go
```diff
@@ -488,6 +490,11 @@ func (m *Manager) CheckIfNeedsRestart(ctx context.Context) (time.Duration, bool)
 
 	robotServiceClient := apppb.NewRobotServiceClient(m.conn)
 	req := &apppb.NeedsRestartRequest{Id: m.cloudConfig.ID}
+	if cs, ok := m.conn.(grpc.ConnectivityState); ok {
+		if cs.GetState() == connectivity.TransientFailure {
```
cheukt: also if it is in connecting state?

cheukt (commented):
the go mod changes seem pretty large for the change we have, but otherwise lgtm

go.mod
aldenh-viam: Yeah it's strange.. I didn't change `go.mod` at all originally but the CI wanted `go mod tidy`: https://github.com/viamrobotics/agent/actions/runs/27647488704/job/81762936948

manager.go
aldenh-viam: When we go offline it's in Connecting briefly, but then stays at Transient Failure, so I figured this just to be conservative. I can add back Connecting though.

go.mod
aldenh-viam: Looks like if I switch to the new `grpchelpers` package from https://github.com/viamrobotics/goutils/pull/571 (`ConnectivityState` moved from rdk's `grpc/` to goutils' `grpchelpers/` ) that gets rid of most of the go.mod/sum diffs besides for version bumps from bumping goutils.

cheukt (approved):

manager.go
```diff
@@ -488,6 +490,11 @@ func (m *Manager) CheckIfNeedsRestart(ctx context.Context) (time.Duration, bool)
 
 	robotServiceClient := apppb.NewRobotServiceClient(m.conn)
 	req := &apppb.NeedsRestartRequest{Id: m.cloudConfig.ID}
+	if cs, ok := grpchelpers.ConnConnectivityState(m.conn); ok {
```
benjirewis: [nit] Could have a helper method for the nested if check and some documentation above it explaining why `TransietFailure` and `connecting` both indicate we should stop whatever scheduled network task we were attempting.

benjirewis (approved):

PR #257, opened 2026-06-17
RSDK-14125 Deduplicate agent logs

benjirewis:
RSDK-14125

## What

Activates noisy-log deduplication in viam-agent and adds support for an agent-specific `disable_log_deduplication` advanced setting, mirroring how the `rdk` does it.

  - Adds a `DisableLogDeduplication` field (disable_log_deduplication JSON key) to AdvancedSettings, along with a `GetDisableLogDeduplication()` wrapper.
  - Threads the logger registry into `Manager` via `NewManager` and adds `applyLogDeduplication`, which flips the registry's `DeduplicateLogs` value based on the config.
  - Applies the setting both at startup (in `NewManager`) and on every cloud config reload (in `GetConfig`), next to the existing `setDebug` call.

## Why

 Log deduplication was never activated for the agent, so noisy logs were emitted without any deduplication. This enables it.

## Testing

  - Added `TestDisableLogDeduplication` in utils/config_test.go: verifies the default is not-disabled, and that true/false parse correctly through the `GetDisableLogDeduplication()` wrapper.
  - Added `TestApplyLogDeduplication` in `manager_test.go`: verifies deduplication is enabled by default, disabled when configured (and re-enabled on a subsequent config update), and that a nil registry is a no-op.

### DIFF
--- a/manager.go
+++ b/manager.go
@@ -65,6 +65,7 @@ type Manager struct {
 	cloudConfig *logging.CloudConfig
 
 	logger      logging.Logger
+	registry    *logging.Registry
 	netAppender *logging.NetAppender
 
 	cfgMu sync.RWMutex
@@ -87,9 +88,16 @@ type Manager struct {
 }
 
 // NewManager returns a new Manager.
-func NewManager(ctx context.Context, logger logging.Logger, cfg utils.AgentConfig, globalCancel context.CancelFunc) *Manager {
+func NewManager(
+	ctx context.Context,
+	logger logging.Logger,
+	registry *logging.Registry,
+	cfg utils.AgentConfig,
+	globalCancel context.CancelFunc,
+) *Manager {
 	manager := &Manager{
 		logger:       logger,
+		registry:     registry,
 		cfg:          cfg,
 		globalCancel: globalCancel,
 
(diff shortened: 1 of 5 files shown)


manager.go
```diff
@@ -398,6 +407,29 @@ func (m *Manager) setDebug(debug bool) {
...
+// Note that the config value is a "disable" while the registry value is an "enable".
+// This is by design to make configuration easier for users and predicates easier for
+// developers respectively. Due to this, the conditional below to check for a diff looks
+// odd (== instead of !=).
+func (m *Manager) applyLogDeduplication(cfg utils.AgentConfig) {
+	if m.registry == nil {
+		return
+	}
+	disable := cfg.AdvancedSettings.GetDisableLogDeduplication()
+	if disable == m.registry.DeduplicateLogs.Load() {
```
aldenh-viam: Would this be clearer? 
```go
shouldDisable := cfg.AdvancedSettings.GetDisableLogDeduplication() // unset tribool returns false here
isDisabled := !m.registry.DeduplicateLogs.Load()
if shouldDisable != isDisabled {
```

or `shouldEnable / isEnabled`?

aldenh-viam (approved):

manager.go
benjirewis: Some better naming here could definitely help. Did something similar.

PR #261, opened 2026-06-22
APP-16389 - check if viam.json exists in msi

cheukt:
image
with CONFIGDIR set
image

### DIFF
--- a/msi/Package.wxs
+++ b/msi/Package.wxs
@@ -16,8 +16,8 @@
 
     <!-- Detect whether the machine config exists at [CONFIGDIR]\viam.json (default C:\etc\viam.json).
          The agent service hangs (retrying forever) when this file is missing, so we fail loudly
-         instead. AppSearch runs early in both the UI and execute sequences -- before the first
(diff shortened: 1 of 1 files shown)


msi/Package.wxs
```diff
@@ -299,6 +324,11 @@
         </Control>
       </Dialog>
 
+      <!-- Check for the machine config the moment the user clicks Next on the Welcome page.
+           When [CONFIGDIR]\viam.json is absent, ViamJsonMissingError shows its message and aborts,
+           so the NewDialog below never runs. -->
```
abe-winter: are you sure you want this to be a hard error rather than a warning? this is better for beginners (makes it very clear what they have to do) and worse for experts (prevents them from installing experimentally on non-configured machines, e.g. VMs in the development flow)

your call -- I'm approving as-is

abe-winter (approved):

msi/Package.wxs
cheukt: i made it a screen - so users can continue if they want, but at least they know. also made it so that it'd fail by default if calling from msiexec /qn but you can FORCE_INSTALL

PR #262, opened 2026-06-29
RSDK-13183 Add a `networking` sublogger for diagnostic logs

benjirewis:
RSDK-13183

## What

Routes all logs from the networking subsystem through a dedicated "networking" sublogger. As a result, every log emitted within the subsystems/networking package now carries the logger name `viam-agent.networking` (and nested names like `viam-agent.networking.network_status`).

## Why

The networking subsystem emits a lot of operational/diagnostic chatter (`NetworkManager` state, connectivity checks, provisioning loop status, etc.) that is not meaningful to most end users in app's "Logs" tab. Those logs can be diagnostic (marked in an app PR here).

### DIFF
--- a/subsystems/networking/networking_linux.go
+++ b/subsystems/networking/networking_linux.go
@@ -76,6 +76,10 @@ type Subsystem struct {
 }
 
 func New(ctx context.Context, logger logging.Logger, cfg utils.AgentConfig) *Subsystem {
+	// Route all networking logs through a "networking" sublogger (e.g. "viam-agent.networking")
+	// so that app can treat them as diagnostic rather than user-facing. See diagnosticLoggerNames
+	// in the app repo.
+	logger = logger.Sublogger("networking")
 	subsys := &Subsystem{
 		cfg:    cfg.NetworkConfiguration,
 		nets:   cfg.AdditionalNetworks,


subsystems/networking/networking_linux.go
```diff
@@ -76,6 +76,10 @@ type Subsystem struct {
 }
 
 func New(ctx context.Context, logger logging.Logger, cfg utils.AgentConfig) *Subsystem {
+	// Route all networking logs through a "networking" sublogger (e.g. "viam-agent.networking")
+	// so that app can treat them as diagnostic rather than user-facing. See diagnosticLoggerNames
+	// in the app repo.
+	logger = logger.Sublogger("networking")
```
benjirewis, line 82: This change routes the entire networking package through the `"viam-agent.networking"` logger, which means all its `Info`/`Warn` logs become diagnostic with [this PR](https://github.com/viamrobotics/app/pull/12558). Reminder that error-level logs are always user-facing. 

I asked Claude if there were any "important" logs that we're now hiding from users. It came up with this candidate list:
- `networking_linux.go:287` — "Both wifi and bluetooth provisioning have been disabled by configuration. Provisioning will not be available."
- `networkmanager_linux.go:730` — "SSID %s connected, but does not provide internet access."
- `networking_linux.go:423` — "Additional networks configured, but internet checking is not enabled. Additional networks may be unused."
- `bluetooth_config_linux.go:39` — "bluetooth version %s is less than 5.50, functionality may be limited"
- `bluetooth_tethering_linux.go:128` — "Failed to complete bluetooth pairing. Please retry the pairing process."
- `bluetooth_tethering_linux.go:125` — "Failed to fully connect to the internet via tethering. Will keep trying. ..."
- `bluetooth_tethering_linux.go:135` — "timed out waiting for bluetooth pairing to complete"
- `bluetooth_loop_linux.go:227` — "BLE startup keeps failing and retries are exhausted. ..."
- `connstate_linux.go:43` — "Online: %t"
- `connstate_linux.go:69` — "Wifi Connected: %t"
- `connstate_linux.go:115` — "Machine credentials present: %t"

The bluetooth tethering state logs could arguably be error-level or logged under a user-facing logger name. But, those logs may never make it to app in the first place since they represent a failure to even establish an internet connection. My take is that what I have now is fine and we can always throw some networking logs under a user-facing log if a user/dev complains.

aldenh-viam (approved):

subsystems/networking/networking_linux.go
cheukt: ya I think there are some connection state/configuration-related logs we'd likely want, but I don't have a good sense of which ones to keep

cheukt (approved):
lgtm

PR #263, opened 2026-07-01
RSDK-14179, RSDK-14180 Agent test quick improvements

EvanDorsky:
Improve docs, do some basic config validation, and make the insecure wifi tests optional (don't run them if no insecure wifi ssid is provided).

### DIFF
--- a/agent-test-example.toml
+++ b/agent-test-example.toml
@@ -25,9 +25,12 @@ viam_agent_old = "<old viam-agent version string e.g. 0.26.0>"
 viam_server_stable = "<current stable viam-server version string e.g. 0.115.0>"
 viam_server_old = "<old viam-server version string e.g. 0.114.0>"
 
+# Provide credentials for a network that the Pi can connect to
+# This network will be used for all tests, including the wifi provisioning tests
 [wifi]
 ssid = "<wifi ssid>"
 password = "<wifi password>"
 
-# leave this blank or commented to skip the insecure wifi tests
-ssid_insecure = "<insecure wifi ssid>"
+# Leave this blank or commented to skip the insecure wifi tests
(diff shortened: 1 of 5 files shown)


agent-test-example.toml
```diff
@@ -6,10 +6,10 @@ api_key = "<app location api key with owner role>"
...
 part_id = "<part id>"
 
-# [serial]
-# serial_path = "<path to serial device, on Linux usually starts with /dev/ttyUSB>"
-# serial_user = "<login username for the serial shell>"
-# serial_pass = "<login password for the serial shell>"
+[serial]
+serial_path = "<path to serial device, on Linux usually starts with /dev/ttyUSB>"
+serial_user = "<login username for the serial shell>"
+serial_pass = "<login password for the serial shell>"
```
aldenh-viam, line 12: Since this and the wifi section were previously commented out, do you want to note whether they're required or not? e.g. will the test suite not run at all if they're left out or left with the placeholder, or will only the tests that require them be skipped?

aldenh-viam (approved):

agent-test-example.toml
EvanDorsky: Yeah, good point. All tests require a serial connection - if the serial config isn't applied, the suite will try to use a default serial path, but the default path doesn't work on macOS (and the connection won't succeed without credentials). So I think it would be better to call these fields required, which would be in line with the comment at the top of this file: `# Uncommented fields are required.`

I think by that logic, though, ~~the wifi fields can be commented out, since those tests are "optional" and there are already hooks to enable/disable them. I'll make that change, hopefully that makes the file more clear.~~

Actually the wifi credentials are required by all tests since they are used to bring the Pi online between test cases. So I'll leave that section how it was.

PR #265, opened 2026-07-06
RSDK-14219 Reduce log verbosity for hotspot provisioning failures

cheukt:
## Problem

The viam-agent logs full stack traces (`errVerbose`) when hotspot provisioning fails due to expected NetworkManager reasons like `NmDeviceStateReasonIpConfigUnavailable`, producing excessively large log entries. Example:

```
ERROR viam-agent networking/networkmanager_linux.go:269 failed to start hotspot provisioning {"err":"...","errVerbose":"connection failed: NmDeviceStateReasonIpConfigUnavailable\ngithub.com/viamrobotics/agent/subsystems/networking.(*Networking).waitForConnect\n\t.../networkmanager_linux.go:504\n... [full stack trace] ..."}
```

The `errVerbose` field is added automatically by zap: when an error value from `github.com/pkg/errors` (aka `errw`) is passed as a structured log field, zap sees it implements `fmt.Formatter` and appends a `${key}Verbose` field containing the full `%+v` stack trace.

## Fix

Pass `err.Error()` instead of the raw error value at the affected structured log sites in the networking subsystem. The wrapped message chain still prints on a single readable line, without the stack blob.

All 7 sites are inside `err != nil` guards, so this is nil-safe (no `.Error()` on nil).

## Scope

Targeted to the networking subsystem — the site named in RSDK-14219 (`startProvisioning` hotspot failure) plus 6 adjacent same-class `Warn` sites (provisioning start/stop, add network, systemctl reboot, rfkill unblock). Log levels and message text are unchanged.

### DIFF
--- a/subsystems/networking/bluetooth_linux.go
+++ b/subsystems/networking/bluetooth_linux.go
@@ -169,7 +169,7 @@ func (n *Subsystem) initializeBluetoothService(
 	ctx context.Context, deviceName string, characteristics []bluetooth.CharacteristicConfig,
 ) error {
 	if err := rfkillUnblock(ctx); err != nil {
-		n.logger.Warnw("Failed to unblock bluetooth with rfkill; bluetooth initialization will continue but may fail", "err", err)
+		n.logger.Warnw("Failed to unblock bluetooth with rfkill; bluetooth initialization will continue but may fail", "err", err.Error())
 	}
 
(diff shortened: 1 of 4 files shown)


benjirewis (approved):
Just a heads up that all the `Warn`-level logs you're editing here will also be diagnostic now.

PR #266, opened 2026-07-07
RSDK-14222 Emit actionable errors on provisioning port conflicts

cheukt:
## Problem

Provisioning could fail with opaque errors when another process holds a port the provisioning stack needs, giving operators no signal about the real cause. Seen in the field (Raspberry Pi OS Trixie): a stray system `dnsmasq` bound to `0.0.0.0:53` blocked NetworkManager's shared-mode dnsmasq from binding `10.42.0.1:53`, surfacing only as `NmDeviceStateReasonIpConfigUnavailable`. Similarly, a service holding `:80` (e.g. nginx wildcard-bound) makes the captive portal fail with a bare "address already in use".

## Change

Enrich these failures **without inspecting the holding process** — just report that the port is taken and hand the operator the one command to find the owner. ~25 LOC, no new files/deps.

- **Portal `:80` / gRPC `:4772`** — the kernel already returns `EADDRINUSE` with the exact port. Special-case it at the `net.Listen` sites in `startWeb`/`startGRPC` and return an enriched error telling the operator to stop the service or scope its listen address (`ss -ltnp | grep :<port>`).
- **Hotspot `:53`** — NM only gives us `IpConfigUnavailable` (not *proven* to be a port conflict — could be a subnet collision). Wrap that reason in a new `ErrIPConfigUnavailable` sentinel in `waitForConnect`, and in `startProvisioningHotspot` log a hedged, actionable hint covering both likely causes (a dnsmasq on `:53`, or a `10.42.0.0/24` collision), with the commands to diagnose and fix.

Deliberately does **not** identify the exact PID/process. For `:80/:4772` there's no loss of precision (`EADDRINUSE` is unambiguous); for `:53` the hedged message also covers the subnet-collision cause a port scan would miss. A follow-up could add best-effort `/proc` process-naming if desired — tracked in the ticket.

## Testing

- `GOOS=linux GOARCH=arm64 go build ./subsystems/networking/...` — clean
- `GOOS=linux GOARCH=arm64 go vet ./subsystems/networking/...` — clean

Jira: RSDK-14222

### DIFF
--- a/subsystems/networking/networkmanager_linux.go
+++ b/subsystems/networking/networkmanager_linux.go
@@ -321,6 +321,13 @@ func (n *Subsystem) startProvisioningHotspot(ctx context.Context) error {
 		return err
 	}
 	if err := n.activateConnection(ctx, n.netState.GenNetKey(NetworkTypeHotspot, "", n.Config().HotspotSSID)); err != nil {
+		if errors.Is(err, ErrIPConfigUnavailable) {
+			n.logger.Error("Hotspot IP configuration failed. This is commonly caused by another " +
+				"DNS/DHCP service (such as a system dnsmasq) holding port 53, which prevents " +
+				"NetworkManager's shared-mode dnsmasq from starting. Check with 'ss -lunp | grep :53' " +
+				"and disable the conflicting service (e.g. 'systemctl disable --now dnsmasq'), or " +
+				"verify nothing else is using the 10.42.0.0/24 subnet.")
+		}
 		return errw.Wrap(err, "starting provisioning mode hotspot")
 	}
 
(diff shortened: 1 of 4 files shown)


subsystems/networking/networkmanager_linux.go
```diff
@@ -321,6 +321,13 @@ func (n *Subsystem) startProvisioningHotspot(ctx context.Context) error {
 		return err
 	}
 	if err := n.activateConnection(ctx, n.netState.GenNetKey(NetworkTypeHotspot, "", n.Config().HotspotSSID)); err != nil {
+		if errors.Is(err, ErrIPConfigUnavailable) {
+			n.logger.Error("Hotspot IP configuration failed. This is commonly caused by another " +
+				"DNS/DHCP service (such as a system dnsmasq) holding port 53, which prevents " +
+				"NetworkManager's shared-mode dnsmasq from starting. Check with 'ss -lunp | grep :53' " +
+				"and disable the conflicting service (e.g. 'systemctl disable --now dnsmasq'), or " +
```
aldenh-viam: is this "e.g." really necessary or loud enough? Don't want people to blindly run this just because it looks like a command.

aldenh-viam (approved):

subsystems/networking/networkmanager_linux.go
cheukt: removed the the e.g.

PR #267, opened 2026-07-07
RSDK-14181: Make managed-security the default for os_auto_upgrade_type

jmatth:
# What

Changes the default value for `os_manageds_upgrade_type` to `managed-security`. This replaces the previous default behavior, which was to do nothing.

# Why

We are migrating agent managed machines to use `managed-security` by default. Updating the way agent treats an empty value in the config is consistent with how other default config values are handled. This PR will need to wait to be merged + deployed until we have finished informing users of the upcoming change and allowing them time to opt out.

### DIFF
--- a/utils/config.go
+++ b/utils/config.go
@@ -503,26 +503,17 @@ func validateConfig(cfg AgentConfig) (AgentConfig, error) {
 		cfg.NetworkConfiguration.HotspotSSID = cfg.NetworkConfiguration.HotspotSSID[:32]
 	}
 
-	var haveBadTimeout bool
 	minTimeout := Timeout(time.Minute)
-	if cfg.NetworkConfiguration.OfflineBeforeStartingHotspotMinutes < minTimeout {
-		//nolint:lll
-		cfg.NetworkConfiguration.OfflineBeforeStartingHotspotMinutes = DefaultConfiguration.NetworkConfiguration.OfflineBeforeStartingHotspotMinutes
-		haveBadTimeout = true
-	}
-
-	if cfg.NetworkConfiguration.UserIdleMinutes < minTimeout {
-		cfg.NetworkConfiguration.UserIdleMinutes = DefaultConfiguration.NetworkConfiguration.UserIdleMinutes
-		haveBadTimeout = true
-	}
-
-	if cfg.NetworkConfiguration.RetryConnectionTimeoutMinutes < minTimeout {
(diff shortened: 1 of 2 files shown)


utils/config_test.go
```diff
@@ -53,7 +53,7 @@ func TestConvertJson(t *testing.T) {
 			"logging_journald_system_max_use_megabytes": 512,
 			"logging_journald_runtime_max_use_megabytes": 512,
 			"logging_journald_storage": "persistent",
-			"os_auto_upgrade_type": "",
+			"os_auto_upgrade_type": "managed-security",
```
aldenh-viam, line 63: Would it be better to leave this as `""` and check if it defaults to managed?

aldenh-viam (approved):

utils/config_test.go
jmatth: Probably, but it would need to go through `validateConfig` where that's actually enforced. I'm actually not sure why this test doesn't already do that. I'll try adding it and see if anything breaks

jmatth: Ok, doing that breaks the test and requires other changes to make it work. I think the way this test is written is just generally really confusing. According to the comment it's just testing that the JSON unmarshaling works, but then it's using a modified copy of the default config to compare against? I'm guessing the default config is used to avoid all the code it would take to construct the expected state manually. I think this should be left as-is since it's just for the JSON behavior, and I'll add a separate test to make sure tha validate method overwrites an empty string with the default

utils/config.go
```diff
@@ -503,26 +503,17 @@ func validateConfig(cfg AgentConfig) (AgentConfig, error) {
...
-		errOut = errors.Join(errOut, errw.New("timeout values cannot be less than 1 minute"))
+	for k, v := range map[string]*Timeout{
+		"offline_before_starting_hotspot_minutes": &cfg.NetworkConfiguration.OfflineBeforeStartingHotspotMinutes,
+		"user_idle_minutes":                       &cfg.NetworkConfiguration.UserIdleMinutes,
+		"retry_connection_timeout_minutes":        &cfg.NetworkConfiguration.RetryConnectionTimeoutMinutes,
+	} {
+		if *v < minTimeout {
+			//nolint:lll
+			*v = DefaultConfiguration.NetworkConfiguration.OfflineBeforeStartingHotspotMinutes
+			errOut = errors.Join(errOut, errors.New(k+": timeout value cannot be less than 1 minute"))
```
aldenh-viam, line 514: perhaps add "set to default x" here

aldenh-viam (approved):

dgottlieb (approved):

PR #269, opened 2026-07-14
RSDK-13903: Improve path handling on Windows

EvanDorsky:
Certain ways of specifying local paths to the agent binary wouldn't work on Windows, and it wasn't clear which kinds of paths should work.

This branch just swaps the usage of `url.Parse` in `DownloadFile` with `urlhelper.Parse` from `go-getter`. On unix, `urlhelper.Parse` literally resolves back to `url.Parse`. On Windows, `urlhelper.Parse` resolves to a custom parser with lots of special casing to handle the various kinds of path problems we were dealing with (it handles drive letters and `file://` url prefixes).

This branch also adds a test with a few different URL schemes for Windows - might be good to add that to CI.

### DIFF
--- a/utils/urlparse_test.go
+++ b/utils/urlparse_test.go
@@ -1,0 +1,67 @@
+package utils
+
+import (
+	"os"
+	"path/filepath"
+	"runtime"
+	"testing"
+
+	"go.viam.com/rdk/logging"
+	"go.viam.com/test"
+)
+
+// TestDownloadFileURLForms documents which URL spellings DownloadFile accepts for a local file
+func TestDownloadFileURLForms(t *testing.T) {
+	MockAndCreateViamDirs(t)
+	logger := logging.NewTestLogger(t)
+
+	const content = "canonical source binary contents"
+	// EvalSymlinks so the path matches what the getter resolves (e.g. /var -> /private/var on darwin).
+	td, err := filepath.EvalSymlinks(t.TempDir())
+	test.That(t, err, test.ShouldBeNil)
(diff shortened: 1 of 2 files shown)


aldenh-viam (approved):
Nice find! Here is go-getter's parse for Windows: https://github.com/viam-labs/go-getter/blob/main/helper/url/url_windows.go

Do you want to add a path without the drive letter (assumes same as cwd), to the test cases? Or leave that as use at your own risk?

utils/urlparse_test.go
```diff
@@ -0,0 +1,71 @@
...
+	vol := filepath.VolumeName(src)
+
+	type urlCase struct {
+		name    string
+		url     string
+		wantErr bool
+	}
+
+	var cases []urlCase
+	if runtime.GOOS == "windows" {
```
jmatth, line 36: I don't see any Window runs on the PR CI. Did you run these manually on a Windows system to confirm they pass?

jmatth (approved):
LGTM with the caveat that we are technically taking on responsibility to support invalid file URIs. I think it would be equally valid to just update the docs with "file URIs must follow this format, here's an example"

utils/urlparse_test.go
EvanDorsky: Yeah, I ran this locally on a Windows machine (and did a bunch of testing on that machine too) and it passed. There isn't really any CI for windows right now but I'm going to take on RSDK-13308 to try to bring in some tests, including this one.

PR #270, opened 2026-07-14
RSDK-13906: Fix agent redownloading itself on fresh install on Windows

EvanDorsky:
This branch adds some logic within agent to check whether the currently-running agent binary is the one that we want to be running, in cases where the version cache doesn't reflect the current-running binary (like after an install that doesn't update the version cache).

It looks like the reason we were redownloading the binary was because the Windows installer didn't update the version cache - but Claude persuaded me that it made more sense to have agent "heal" this case itself, since there are also certain paths in the unix install that can put agent in the same state.

Also added a test for this logic. Now that this branch is up, I think I can do an end-to-end test by building the dev release binary and pinning a machine to that, then installing on Windows with a custom MSI.


cheukt (approved):
lgtm if manually testing looks good

PR #279, opened 2026-07-29
RSDK-14344: Don't reboot while an OS package upgrade is installing

cheukt:
Fixes [RSDK-14344](https://viam.atlassian.net/browse/RSDK-14344).

## Problem

In managed upgrade modes the agent both installs OS updates and owns the reboot, but the two loops share no state. The syscfg upgrade worker runs `apt-get`/`dnf`/`Get-WindowsUpdate -Install` for minutes, while the reboot poller (`manager.go:730-745`) checks every 60s and reboots if one is pending and the maintenance window is open. Nothing tells the poller an install is in flight, so a reboot can land mid-transaction — a dpkg database needing `dpkg --configure -a` on Linux, or a rollback loop on Windows.

This isn't a narrow interleaving: `manager.go:156` passes `viamServer.RestartAllowed` as syscfg's `maintenanceAllowed`, so "upgrade may run" and "reboot may fire" are the same predicate. Two overlap paths need no timing luck — a config change re-runs `Stop`/`Update`/`Start` with `needsOSReboot` still latched, and after a process restart the agent re-derives "reboot pending" from the on-disk indicator while the startup install runs. On Linux that indicator is set mid-transaction anyway: `/var/run/reboot-required` is written by package **postinst** scripts (`libc6`, `libpam0g`, `libssl3` on stock Ubuntu 22.04) during `dpkg --configure`.

## Fix

Gate the reboot on two checks, since OS state outlives agent state:

- **In-process flag** (`upgradeState`) around our own install. Carries its own mutex rather than reusing `s.mu`, which `Stop` holds while waiting for the upgrade worker to exit.
- **OS-level probe** for transactions the agent didn't start — `F_GETLK` against the dpkg/rpm lock files on Linux (non-destructive: it queries the holder, acquires nothing), and on Windows `IUpdateInstaller.IsBusy` OR'd with the TrustedInstaller service state. The two Windows signals cover different servicing paths and neither subsumes the other: `IsBusy` sees WUA install sessions, including MSI- and driver-shaped updates that never touch the component store, while TrustedInstaller sees component-based servicing from any source (`.msu`, DISM, feature changes) and keeps running after the WUA session that queued the work has ended. Unreadable lock state counts as busy on Linux; on Windows both probes failing reads as *not* busy, so a host with broken PowerShell still reboots rather than never rebooting.

Both `NeedsOSReboot` implementations now separate *is a reboot pending* (latched) from *is it safe to act now* (re-evaluated each poll), with the gate applied to the cached-true path too. A held-back reboot logs once via `blockNotice` and rearms when the condition clears.

## Testing

- Cross-compiles for linux/windows/darwin; `mise run lint` clean for `GOOS=linux` and `GOOS=windows`.
- Full `go test -race` passes on Linux. (`utils.TestInitPaths/failure_cannot_create_directory` fails when the suite runs as root in a container — pre-existing, passes as non-root.)
- `TestPackageLockHeld` re-execs the test binary as a child holding a real `F_SETLK` write lock (`F_GETLK` ignores the caller's own locks), asserts the probe reports held with that PID, and re-probes to confirm it didn't take the lock. `TestNeedsOSRebootWaitsForUpgradeToFinish` covers the end-to-end suppress/release.

## Reviewer notes

- **Windows paths are compile- and lint-verified here** — no Windows host on my end to exercise `IsBusy` against a live update session; @EvanDorsky tested on Windows, which is where the OR'd TrustedInstaller check came from.
- Known cost of the Windows gate: TrustedInstaller lingers idle for some minutes after servicing finishes, and `CheckIfOSNeedsReboot` needs both "not busy" and an open maintenance window. A linger that outlasts a short window defers the reboot to the next one. Accepted over force-rebooting a transaction we can't see the end of.
- Open question tracked as [RSDK-14345](https://viam.atlassian.net/browse/RSDK-14345), not a blocker: whether the WUA writes `RebootRequired` per-update or at session end. The `IsBusy` gate holds either way.
- Not addressed: cancelling the upgrade context SIGKILLs the package manager (`pkgCmd` uses `exec.CommandContext`), so agent shutdown can kill `apt-get` mid-transaction with no reboot involved. Same corruption class, different trigger; noted on RSDK-14344.

[RSDK-14344]: https://viam.atlassian.net/browse/RSDK-14344?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[RSDK-14345]: https://viam.atlassian.net/browse/RSDK-14345?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/subsystems/syscfg/managed_upgrades_locks_linux.go
+++ b/subsystems/syscfg/managed_upgrades_locks_linux.go
@@ -6,7 +6,6 @@ package syscfg
 // locks live in the kernel and do.
 
 import (
-	"context"
 	"io"
 	"os"
 
@@ -15,23 +14,11 @@ import (
 	"golang.org/x/sys/unix"
 )
 
-// packageLockPaths are the lock files package managers hold for the duration of
-// a transaction. A variable so tests can substitute paths they control.
-var packageLockPaths = []string{
-	// Held across the whole transaction including spawned dpkg calls, so it is the
-	// broadest signal on Debian-likes.
-	"/var/lib/dpkg/lock-frontend",
-	"/var/lib/dpkg/lock",
-	// Covers both dnf and yum.
-	"/var/lib/rpm/.rpm.lock",
-}
-
-// osUpgradeInProgress reports whether any package manager currently holds a
-// transaction lock. A missing lock file just means that package manager is not
-// installed. An unreadable one reads as not-held, so that a permanently failing
-// check cannot block reboots forever.
-func osUpgradeInProgress(logger logging.Logger) bool {
-	for _, path := range packageLockPaths {
+// osUpgradeInProgress reports whether another process holds any of paths as a
+// transaction lock. Missing and unreadable files both read as not-held, so that a
+// permanently failing check cannot block reboots forever.
+func osUpgradeInProgress(logger logging.Logger, paths []string) bool {
+	for _, path := range paths {
 		held, pid, err := packageLockHeld(path)
 		switch {
 		case err != nil:
@@ -55,7 +42,7 @@ func osUpgradeInProgress(logger logging.Logger) bool {
 // always a child process, already covered by upgradeState.
 func packageLockHeld(path string) (bool, int32, error) {
 	// Not O_CREATE: we must never create a lock file a package manager would later
-	// rely on. Paths come from packageLockPaths, never from config.
+	// rely on. Paths are compiled in, never from config.
 	//nolint: gosec
 	f, err := os.Open(path)
 	if err != nil {
@@ -83,8 +70,8 @@ func packageLockHeld(path string) (bool, int32, error) {
 
 // rebootBlockedByOSUpgrade reports whether a pending reboot must be held back
 // because a package transaction is running, logging the first time it does.
-func (s *Subsystem) rebootBlockedByOSUpgrade(_ context.Context) bool {
-	blocked := osUpgradeInProgress(s.logger)
+func (s *Subsystem) rebootBlockedByOSUpgrade(pkgMgr packageManager) bool {
+	blocked := osUpgradeInProgress(s.logger, pkgMgr.lockPaths())
 	s.rebootBlocked.note(s.logger, blocked,
 		"OS reboot pending, but a package manager transaction is in progress; waiting for it to finish")
 	return blocked
--- a/subsystems/syscfg/managed_upgrades_locks_linux.go
+++ b/subsystems/syscfg/managed_upgrades_locks_linux.go
@@ -1,0 +1,95 @@
+package syscfg
+
+// This file probes whether a package manager transaction is running anywhere on
+// the system, including one the agent did not start (a concurrent
+// unattended-upgrades run, or an operator at a shell). The in-process
+// upgradeState flag only covers upgrades this process is executing, and it does
+// not survive an agent restart; these locks live in the kernel and do.
+
+import (
+	"context"
+	"io"
+	"os"
+
+	"go.viam.com/rdk/logging"
+	goutils "go.viam.com/utils"
+	"golang.org/x/sys/unix"
+)
+
+// packageLockPaths are the lock files package managers hold for the duration of
+// a transaction. A variable rather than a constant so tests can substitute
+// paths they control.
+var packageLockPaths = []string{
+	// APT holds the frontend lock across an entire transaction, including the
+	// dpkg invocations it spawns, so it is the broadest signal on Debian-likes.
+	"/var/lib/dpkg/lock-frontend",
+	// dpkg's own lock, held while unpacking and configuring packages.
+	"/var/lib/dpkg/lock",
+	// RPM's transaction lock, covering both dnf and yum.
+	"/var/lib/rpm/.rpm.lock",
+}
+
+// osUpgradeInProgress reports whether any package manager currently holds a
+// transaction lock. Missing lock files (RPM's on a Debian system, and vice
+// versa) are not an error, they just mean that package manager is not present.
+func osUpgradeInProgress(logger logging.Logger) bool {
+	for _, path := range packageLockPaths {
+		held, pid, err := packageLockHeld(path)
+		switch {
+		case err != nil:
+			if !os.IsNotExist(err) {
+				logger.Debugw("Could not check package manager lock", "path", path, "err", err)
+			}
+		case held:
+			logger.Debugw("Package manager transaction in progress", "path", path, "pid", pid)
+			return true
+		}
+	}
+	return false
+}
+
+// packageLockHeld reports whether another process holds a write lock on path,
+// along with the PID holding it.
+//
+// This uses F_GETLK rather than trying to take the lock: it asks the kernel
+// whether a lock would conflict and returns the holder, without acquiring
+// anything, so it cannot interfere with the transaction it is inspecting. Note
+// that F_GETLK reports only locks held by *other* processes, which is what we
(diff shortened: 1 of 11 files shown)


subsystems/syscfg/managed_upgrades_windows.go
```diff
@@ -171,6 +197,44 @@ func windowsRebootRequired(ctx context.Context) bool {
...
+// servicing finishes and so can delay a reboot, but treating an unreadable state
+// as "not busy" would reopen the race this guards against.
+//
+// Note that IsBusy only covers the Windows Update Agent; a bare MSI or a vendor
+// driver installer running outside it is not visible here.
+func windowsUpdateBusy(ctx context.Context) bool {
+	out, err := exec.CommandContext(ctx, "powershell",
+		"-NonInteractive", "-NoProfile",
+		"-Command", `(New-Object -ComObject Microsoft.Update.Installer).IsBusy`,
+	).Output()
```
EvanDorsky, line 222: Tested on Windows - going back and forth with Claude, it sounds like it might be better to OR the `IsBusy` check with the `TrustedInstaller` check below, instead of only falling back to `TrustedInstaller` after an `IsBusy` call fails.

Reasoning is it looks like the two signals report on different things (different types of updates), all of which we want to be responsive to. It looks like the only cost would be reboots delayed by a few minutes since the signal from `TrustedInstaller` can linger after updates have completed - which doesn't sound that bad.

subsystems/syscfg/managed_upgrades_windows.go
```diff
@@ -104,7 +119,14 @@ func (s *Subsystem) runManagedUpgrade(ctx context.Context) error {
 
 	s.logger.Info("Running managed Windows Update")
 
-	if err := runWindowsUpdate(ctx, mode == utils.OSAutoUpgradeManagedSecurity); err != nil {
+	// Block reboots for the duration of the install: Windows sets RebootRequired as
+	// soon as one update needs it, so the key can appear while the rest of the batch
+	// is still being written to the component store.
+	err := func() error {
+		defer s.upgrade.begin()()
```
jmatth: [nit] At first glance this line looks like it's deferring starting the upgrade. Might be a bit clearer to create an intermediate variable to make clear we're starting the upgrade and then deferring clearing the flag.

```suggestion
		upgradeDone := s.upgrade.begin()
		defer upgradeDone()
```

subsystems/syscfg/managed_upgrades_locks_linux.go
```diff
@@ -0,0 +1,91 @@
...
+// packageLockPaths are the lock files package managers hold for the duration of
+// a transaction. A variable so tests can substitute paths they control.
+var packageLockPaths = []string{
+	// Held across the whole transaction including spawned dpkg calls, so it is the
+	// broadest signal on Debian-likes.
+	"/var/lib/dpkg/lock-frontend",
+	"/var/lib/dpkg/lock",
+	// Covers both dnf and yum.
+	"/var/lib/rpm/.rpm.lock",
+}
```
jmatth: Personally I'd move these into the `apt` and `rpm` specific implementations as a method on the `packageManager` interface

subsystems/syscfg/managed_upgrades_locks_linux.go
```diff
@@ -0,0 +1,91 @@
...
+
+// packageLockPaths are the lock files package managers hold for the duration of
+// a transaction. A variable so tests can substitute paths they control.
+var packageLockPaths = []string{
+	// Held across the whole transaction including spawned dpkg calls, so it is the
+	// broadest signal on Debian-likes.
+	"/var/lib/dpkg/lock-frontend",
+	"/var/lib/dpkg/lock",
+	// Covers both dnf and yum.
+	"/var/lib/rpm/.rpm.lock",
```
jmatth: From what I can find this file location changed in RHEL 9. If we don't care about RHEL 8 and older then this is fine, otherwise we need to check if any processes are using the Berkely DB files `/var/lib/rpm/__db.*`.

Theoretically this also might leave the package cache in a bad state? As far as I can tell this file is just for rpm. dnf/yum might be performing other work like downloading package files though, and a sudden reboot could leave partially downloaded files in the cache. Their lockfile locations are even less well documented though and at least dnf doesn't provide a nice way to ask if the lock is held via the cli, so I'm inclined to say that's an edge case we won't address  and it's up to the package manager to identify and reacquire incomplete or corrupted package files.

jmatth (approved):
Left a couple comments but broadly I think this makes sense. I'm on the fence whether it's worth it to add and maintain code that both blocks reboots when we're installing a managed update _and_ checks whether something else is performing an update. It seems like we're repeating work but they can break in different ways so I guess it's worth it to be paranoid here.

subsystems/syscfg/managed_upgrades_locks_linux.go
cheukt: Claude here — I went and checked this in containers rather than answering from memory, and I don't think the RHEL 9 change applies: `/var/lib/rpm/.rpm.lock` is stable back to RHEL 7.

`.rpm.lock` is the `%_rpmlock_path` macro, and it has been rpm's fcntl transaction lock since ~4.4 (2006), guarding `rpmtsRun()` / `rpmtsInitDB()` / `rpmtsRebuildDB()` / `rpmtsVerifyDB()`. The BDB→sqlite switch in rpm 4.16 (RHEL 9) changed where package *data* lives — `Packages` became `rpmdb.sqlite` — but not the lock. `__db.*` are BDB *environment* files, not the transaction lock.

| image | rpm | `rpm --eval %{_rpmlock_path}` | file exists | backend |
|---|---|---|---|---|
| `centos:7` | 4.11.3 | `/var/lib/rpm/.rpm.lock` | yes | BDB |
| `rockylinux:8` | 4.14.3 | `/var/lib/rpm/.rpm.lock` | yes | BDB (`Packages` present) |

I also confirmed the probe fires against a real transaction on Rocky 8, not just that the file is there — an `F_GETLK` script mirroring `packageLockHeld` polling at 20ms while `dnf -y install vim-enhanced httpd git` ran:

```
HELD type=1 pid=88
RESULT: lock observed held = True
```

`type=1` is `F_WRLCK` and `pid=88` was the dnf process; it read unlocked before the transaction and after. So the existing path covers RHEL 8 on the BDB backend and no `__db.*` fallback is needed.

Agreed on the dnf/yum cache point, and leaving it alone. The window we actually care about is `rpmtsRun`, which `.rpm.lock` covers; a partially downloaded package is rpm's own to re-fetch and doesn't need `dpkg --configure -a`-class recovery.

Happy to add a one-line comment recording that the path is `%_rpmlock_path` and is backend-independent, if it's worth documenting so this doesn't resurface.

subsystems/syscfg/managed_upgrades_windows.go
cheukt: updated

subsystems/syscfg/managed_upgrades_locks_linux.go
cheukt: makes sense, done

subsystems/syscfg/managed_upgrades_windows.go
cheukt: added

PR #282, opened 2026-07-30
Verify published manifests match the binaries they point at

cheukt:
## Problem

#280 stopped a release-tagged commit from publishing prerelease artifacts, closing the route by which `viam-agent-v1.0.1-windows-x86_64.json` came to advertise a sha256 no released binary had. But nothing *detects* a manifest/binary disagreement, so if one arises by another route it stays invisible until Windows machines start rejecting downloads and retrying forever.

## What this adds

**`verify-manifests.sh`** reads a published manifest's `sha256`, downloads whatever object its own `upload-path` names, and compares. It never consults a local build — it answers "is the bucket self-consistent?" rather than "does the bucket match me?", which makes it meaningful regardless of which of two racing runs published last. Published reads are anonymous; no credentials except under `--fix`.

```
./verify-manifests.sh --version v1.0.1   # every platform for one version
./verify-manifests.sh --count 20         # the 20 most recently written manifests
./verify-manifests.sh --count 0          # every release manifest
```

**Test and Build** runs it against the version it just published. **Verify Manifests** (dispatch) sweeps on demand. **Fix Manifest** (dispatch) repairs a mismatch in whichever direction is correct.

## Why the check is gated on step outcome

```yaml
if: steps.upload_release.outcome == 'success' || steps.upload_prerelease.outcome == 'success'
```

rather than on a copy of the upload steps' `if` conditions. Restating `contains(version, '-dev.')` would tie the check to the same predicate that guards the prerelease upload, so a regression in that guard would skip the check in exactly the run that just clobbered a manifest.

Keying off what actually ran means any run that publishes also re-reads what it published, so the last writer in a race always inspects the state it left behind. Replaying the original incident, the release run's verify reads the bucket tens of seconds after its upload — well past the losing run's write at +1.6s — and fails on the windows manifest.

## Why Fix Manifest has two modes

A mismatch looks identical whichever side is at fault, and the script has no information to choose:

- **`manifest`** — the published binary is what you meant to ship and only the checksum is wrong. Rewrites the manifest; binaries untouched. This is the repair for a concurrent-run clobber, and the one v1.0.1 needs.
- **`rebuild`** — the binary itself is wrong. Rebuilds the tag, re-signs, replaces binary + manifest + MSI from one build. Windows binaries aren't reproducible (Authenticode timestamps), so this publishes new bytes rather than restoring the old ones.

Applying the wrong one is a one-way door: rewriting the manifest when the *binary* is wrong permanently blesses the wrong artifact, since the checksum then validates. So the script prints the corrected manifest and the operator picks; nothing passes `--fix` automatically.

`rebuild` deliberately does not touch the `stable` aliases — `stable` tracks the newest release, not the tag being repaired.

## Test

Verified against the live production bucket:

| Check | Result |
|---|---|
| `--version v1.0.1` | flags windows, passes the other 3 platforms |
| `--version v1.0.2` | all 4 match |
| `--version v1.0.1-dev.1` | all 4 match (prerelease paths) |
| `--count` sweep | works, ~55s / ~600MB for 20 |

`shfmt -d *.sh` clean, all workflows parse. The dispatch workflows are not exercised end-to-end — they need production bucket credentials.

## Follow-ups, not fixed here

- **v1.0.1's manifest is still wrong in production.** Needs Fix Manifest with `mode=manifest`; its sha should be `8e0c004af9f156f4bdef1ca9cff0bb0544653c17706660deb742909bbd139fc4`.
- **Neither repair unsticks already-affected machines.** `version_control.go:182` returns before assigning `UnpackedSHA` on line 197, so a corrected checksum for an unchanged version string is never re-read.
- **PR dev-release manifests have a broken `upload-path`.** `manifest.sh`'s version regex doesn't recognise `-pr.`, so PR manifests get a release-path `upload-path` while their binaries go to `prerelease/pr-N/`. Same underlying flaw as the v1.0.1 bug, still latent — PR runs are excluded from the check for this reason.
- **Neither dispatch workflow has an approval gate** beyond a typed confirmation. If there's a GitHub environment with required reviewers, putting `environment:` on the writing jobs would be stronger.

### DIFF
--- a/.github/workflows/fix-manifest.yaml
+++ b/.github/workflows/fix-manifest.yaml
@@ -1,0 +1,180 @@
+name: Fix Manifest
+
+# Repairs a version whose published manifest and binary disagree, as reported by
+# Test and Build's "Verify manifests published by this run" step, by the Verify
+# Manifests workflow, or by running verify-manifests.sh directly.
+#
+# There are two repairs and they are not interchangeable, because a mismatch
+# looks identical whichever side is at fault. Pick by deciding what the object
+# currently at upload-path is:
+#
+#   manifest  The published binary is the artifact you meant to ship, and only
+#             the manifest's sha256 is wrong. Rewrites the manifest to match the
+#             binary. Published binaries are untouched. This is the repair for
+#             the concurrent-run clobber, where both runs built the same source
+#             and the release binary is correctly signed.
+#
+#   rebuild   The binary itself is wrong, so its checksum should not be blessed.
+#             Rebuilds the tag from source, re-signs, and re-uploads binary,
+#             manifest and MSI from one build so they cannot disagree.
+#             This REPLACES published artifacts. Windows binaries are not
+#             reproducible (Authenticode timestamps), so the rebuilt exe and MSI
+#             will differ byte-for-byte from what was released, even though the
+#             source is identical. Linux and darwin rebuild byte-identical under
+#             -trimpath.
+#
+# Neither repair unsticks machines that already cached the bad sha:
+# version_control.go returns early on an unchanged version string, so it never
+# re-reads a corrected checksum. Those need a version bump or a cache clear.
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'Version tag to repair, e.g. v1.0.1'
+        required: true
+        type: string
+      mode:
+        description: 'manifest = trust the published binary, fix the manifest. rebuild = rebuild the tag and replace the artifacts.'
+        required: true
+        type: choice
+        options:
+          - manifest
+          - rebuild
+      confirm:
+        description: 'Re-type the version tag to confirm. Both modes write to the production bucket.'
+        required: true
+        type: string
+
+permissions:
+  contents: read
+
+jobs:
+  # Cheap and separate so a typo fails before anything authenticates to GCS.
+  guard:
+    name: Confirm target
+    runs-on: ubuntu-latest
+    steps:
+    - name: Check confirmation matches
+      run: |
+        if [ "${{ inputs.confirm }}" != "${{ inputs.version }}" ]; then
+          echo "confirm (${{ inputs.confirm }}) does not match version (${{ inputs.version }})" >&2
+          exit 1
+        fi
+        case "${{ inputs.version }}" in
+          v[0-9]*.[0-9]*.[0-9]*) ;;
+          *) echo "version must look like v1.0.1" >&2; exit 1 ;;
+        esac
+
+  fix-manifest:
+    name: Rewrite manifest to match published binary
+    needs: guard
+    if: inputs.mode == 'manifest'
+    runs-on: ubuntu-latest
+    permissions:
+      contents: 'read'
+      id-token: 'write'
+    steps:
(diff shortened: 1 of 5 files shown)


.github/workflows/workflow.yaml
```diff
@@ -141,10 +142,26 @@ jobs:
...
     - name: Upload binaries to GCS (Prerelease)
+      id: upload_prerelease
       if: github.event_name != 'release' && github.event_name != 'pull_request' && contains(steps.version.outputs.value, '-dev.')
       run: |
         gsutil -h "Cache-Control:no-cache" cp bin/viam-agent-* gs://packages.viam.com/apps/viam-agent/prerelease/
         gsutil -h "Cache-Control:no-cache" cp etc/viam-agent-*.json gs://packages.viam.com/apps/viam-subsystems/
+    # Every manifest this run wrote has to advertise the sha256 of the object its
+    # upload-path names. Because this runs after the upload, any run that writes
+    # a manifest also re-reads it afterwards, so the last writer in a race always
+    # inspects the state it left behind. That is what makes detection independent
```
danielbotros, line 153: Q: I know there are some `concurrency` configurations gh actions provides, is there a way we can use those to make a race not possible or less damaging? Can be a follow up.

verify-manifests.sh
```diff
@@ -0,0 +1,255 @@
...
+		exit 0
+		;;
+	*)
+		echo "unknown argument: $1" >&2
+		usage >&2
+		exit 2
+		;;
+	esac
+	shift
+done
```
danielbotros, line 69: Should we make `--count` and `--fix` mutually exclusive?

.github/workflows/fix-manifest.yaml
```diff
@@ -0,0 +1,180 @@
...
+      with:
+        ref: ${{ inputs.version }}
+    - uses: actions/download-artifact@v4
+      with:
+        name: viam-agent-windows
+        path: msi
+    - name: Build MSI
+      shell: bash
+      run: |
+        mv msi/viam-agent-${{ inputs.version }}-windows-x86_64 msi/viam-agent-from-installer.exe
```
danielbotros, line 163: Totally not worth the `workflow.yaml` refactor right now, but I wonder if we can re-use steps from that main job here so there's less surface area for things like the `viam-agent-from-installer.exe` filename to drift, if it ever changes.

danielbotros (approved):
LGTM! Just some comments about future follow ups.

verify-manifests.sh
cheukt: made --fix require --version

.github/workflows/workflow.yaml
cheukt: we definitely could - though in theory https://github.com/viamrobotics/agent/commit/2d56af07d2960ef31fb4f120bf664f84a31cbc1a should more or less prevent this from happening again. but good to have more checks

.github/workflows/fix-manifest.yaml
cheukt: Agreed, and filed as https://viam.atlassian.net/browse/RSDK-14389 so the duplication does not get lost.

PR #286, opened 2026-08-04
Add more detailed os update logging

jmatth:
This PR adds logging around os auto updates, primarily by logging which updates will be installed just before applying them, and then which updates were installed (or failed to install) immediately after. Much of the code is dedicated to parsing the package manager output into go structs that can be logged as zap fields to allow for easier db queries in the future if we need to diagnose an issue or set up log based alerts. If parsing fails for whatever reason we log a warning but continue installing updates.

This also uses the new activity logs feature from rdk to log specific events around os updates. The activity log for triggering a reboot in particularly could be used as the basis for a log based alert in the future.

During testing I observed some interesting behavior that I wasn't able to root cause but is probably worth mentioning

- A Windows update was killed mid application. The timeline was successful Windows update installation -> reboot -> agent starts installing more updates -> updates aborted as agent shuts down -> agent starts intsalling more updates -> updates complete. My best guess is that a major Windows update was installed that required multiple automatic reboots, and as a Windows service agent was able to start just long enough to start working before the second reboot hit and it had to shut down

- Alma Linux (RHEL) consistently logs a failure to reboot a couple times before succeeding. Apparently RHEL includes a mechanism for certain tasks to block `systemctl reboot`, even by root. This actually seems like a pretty nice feature, but the logs are potentially confusing

### DIFF
--- a/manager.go
+++ b/manager.go
@@ -1017,13 +1017,11 @@ func (m *Manager) CheckIfOSNeedsReboot(ctx context.Context) {
 	} else {
 		rebootCmd = exec.CommandContext(ctx, "systemctl", "reboot")
 	}
-	m.logger.Activity(syscfg.UpdateActivity, "reboot")
+	m.logger.Activity("system", "reboot", "reason", "os_update")
 	if output, err := rebootCmd.CombinedOutput(); err != nil {
-		m.logger.Activity(syscfg.UpdateActivity, "reboot_abort", "error", err, "output", string(output))
+		m.logger.Activity("system", "reboot_abort", "error", err, "output", string(output))
 		return
 	}
-
-	m.logger.Activity("system", "reboot", "reason", "os_update")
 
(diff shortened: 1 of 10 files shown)


EvanDorsky:
I'm basically trusting your methodology here and just grepping for the new log lines + activity logs.

One thing is, there are some existing activity logs for reboots requested by agent that are labeled `"system"`, and this branch introduces a new `"os update"` category.

I think it would be better to put all these reboots under the `"system"` category so that a reboot is a reboot, and then use reasons to tell os update reboots apart from other kinds of reboots. It looks like this could just be a matter of changing `UpdateActivity` to `"system"` and maybe adding some more metadata in a few places (I'm still reviewing but will leave comments on anything I find).

Other than that, this looks good.

EvanDorsky (approved):
lgtm with the logs moved under `system`!

cheukt (approved):
lgtm - tested on a debian vm and the new logs are great

PR #288, opened 2026-08-17
RSDK-13308: Windows CI for agent

EvanDorsky:
Pull some already-passing tests into Windows CI just to get some coverage going.

There's future work to actually bring the suites up to parity but this is a start.


.github/workflows/workflow.yaml
```diff
@@ -70,6 +70,49 @@ jobs:
...
+    name: Test on Windows
+    runs-on: windows-2022
+    if: |
+      github.event_name != 'pull_request' ||
+      (github.event.action != 'labeled' && github.event.action != 'unlabeled') ||
+      (github.event.action == 'labeled' && github.event.label.name == 'dev-release')
+    steps:
+    - uses: actions/checkout@v4
+      with:
+        fetch-depth: 0 # Required to be able to generate prerelease tags
```
jmatth: Do we still need this just to run the tests?

jmatth (approved):
I <3 more test coverage

.github/workflows/workflow.yaml
EvanDorsky: Good catch, no, doesn't appear necessary.

PR #291, opened 2026-08-24
Fix managed updates issues on Ubuntu

jmatth:
Systems that use `apt` for package management need to use the separate `unattended-upgrades` program to filter updates to apply only security updates. We were using a filter config specific to Debian, which caused `managed-security` to install no updates on Ubuntu. This PR updates the config to work on both systems.

There was also a related issue where we failed to disable the background job that gets installed along with the `unattended-upgrades` program because that code was gated behind the same check we use only enable unettended-upgrades on supported debian versions. That is resolved by moving the disable logic before the check.

During testing I also noticed that we would emit an activity log noting the end of os updates even when we never logged a _start_ to the update because no updates were available. I added an early return to that code path so update end activity logs now always follow an update start activity log.

Tested on Debian and Ubuntu VMs

### DIFF
--- a/subsystems/syscfg/managed_upgrades_windows.go
+++ b/subsystems/syscfg/managed_upgrades_windows.go
@@ -142,6 +142,8 @@ func (s *Subsystem) runManagedUpgrade(ctx context.Context) error {
 	pending, listErr := pendingWindowsUpdates(ctx, s.logger, securityOnly)
 	updates := updateSummary{updates: pending, listErr: listErr}
 	logPendingUpdates(s.logger, updates, "security_only", securityOnly)
+	// Bail early if we were able to parse the update check output and it reports
+	// there are no updates to install.
 	if !updates.shouldContinue() {
 		return nil
 	}
--- a/subsystems/syscfg/managed_upgrades_windows.go
+++ b/subsystems/syscfg/managed_upgrades_windows.go
@@ -142,6 +142,9 @@ func (s *Subsystem) runManagedUpgrade(ctx context.Context) error {
 	pending, listErr := pendingWindowsUpdates(ctx, s.logger, securityOnly)
 	updates := updateSummary{updates: pending, listErr: listErr}
 	logPendingUpdates(s.logger, updates, "security_only", securityOnly)
+	if !updates.shouldContinue() {
+		return nil
+	}
 
 	installed, upgradeErr := installWindowsUpdates(ctx, s.logger, securityOnly)
 	logUpgradeResult(ctx, s.logger, installed, upgradeErr, "security_only", securityOnly)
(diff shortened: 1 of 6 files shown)


subsystems/syscfg/managed_upgrades_windows.go
```diff
@@ -142,6 +142,9 @@ func (s *Subsystem) runManagedUpgrade(ctx context.Context) error {
 	pending, listErr := pendingWindowsUpdates(ctx, s.logger, securityOnly)
 	updates := updateSummary{updates: pending, listErr: listErr}
 	logPendingUpdates(s.logger, updates, "security_only", securityOnly)
+	if !updates.shouldContinue() {
```
EvanDorsky: nit: It's not clear to me why we make this check here and not somewhere else, or what would cause this condition to be met, I think a comment could help.

subsystems/syscfg/upgrades_linux.go
```diff
@@ -69,6 +71,15 @@ func (s *Subsystem) EnforceUpgrades(ctx context.Context) error {
 		return nil
 	}
 
+	unsupportedReason, err := checkSupportedDistro()
+	if err != nil {
+		return err
+	}
+	if unsupportedReason != "" {
+		s.logger.Infow("Skipping unattended upgrades configuration", "reason", unsupportedReason)
```
EvanDorsky: should this be warn?

subsystems/syscfg/upgrades_linux.go
```diff
@@ -188,20 +199,23 @@ func generateOrigins(ctx context.Context, securityOnly bool) (string, error) {
 
 // inner transformation logic of generateOrigins for testing.
 func generateOriginsInner(securityOnly bool, output []byte) map[string]bool {
-	releaseRegex := regexp.MustCompile(`release.*o=([^,]+).*n=([^,]+).*`)
+	// Match the archive/suite (a=) rather than the codename (n=): Ubuntu keeps
+	// the pocket in Suite (a=jammy-security, n=jammy) while Debian carries it in
```
EvanDorsky: maybe this is a term I should know but I have no idea what "pocket" refers to here

EvanDorsky (approved):
lgtm, just some nits on comments

subsystems/syscfg/upgrades_linux.go
jmatth: Apparently it's mostly used by Ubuntu. [Here's an example][1]. I'll update the comment to be less obtuse though

[1]: https://wiki.ubuntu.com/SecurityTeam/FAQ#How_are_the_.22-updates.22_and_.22-security.22_pockets_different.3F

subsystems/syscfg/upgrades_linux.go
jmatth: Possibly, I just didn't want to mess with existing log lines but I can change it

